PR #17, opened 2015-03-16
Features/ingestform

### LINKED ISSUE #3: Ingest form
As an administrator I want to be able to upload a single file and associate RDF metadata with the object so that it can be displayed to users.

zhang4952:
fixes #3


app/presenters/my_generic_file_presenter.rb
```diff
@@ -0,0 +1,6 @@
+# app/presenters/my_generic_file_presenter.rb
+class MyGenericFilePresenter < Sufia::GenericFilePresenter
```
terrellt: Can we rename this to FilePresenter?

app/forms/my_file_edit_form.rb
```diff
@@ -0,0 +1,7 @@
+# app/forms/my_file_edit_form.rb
+class MyFileEditForm < MyGenericFilePresenter
```
terrellt: Rename to FileEditForm?

app/forms/my_batch_edit_form.rb
```diff
@@ -0,0 +1,3 @@
+# app/forms/my_batch_edit_form.rb
+class MyBatchEditForm < MyFileEditForm
```
terrellt: Rename to BatchEditFOrm?

app/presenters/file_presenter.rb
```diff
@@ -0,0 +1,6 @@
+# app/presenters/file_presenter.rb
+class FilePresenter < Sufia::GenericFilePresenter
+  self.terms = [:resource_type, :title, :creator, :contributor, :description,
```
terrellt: Not all the properties in the model are here - is that on purpose?

terrellt:
Branch needs rebased.

straleyb:
Looks good. @luisgreg99?

terrellt:
@zhang4952 Can you squash these commits? This is a lot of noise in the commit history.

straleyb:
@zhang4952 @terrellt Is a generic file supposed to have an attribute called "type"?

terrellt:
Can't have type, it's a reserved term.

straleyb:
Looks like it needs to be taken out of the presenter then.

luisgreg99:
@zhang4952 , I got an undefined attribute error after uploading a file as an admin. @straleyb suggested removing the "type" field in the FilePresenters class, which fixes the error. Is the "type" field required?

luisgreg99:
In that case, we can rename it if it's a required field.

luisgreg99:
@zhang4952 I tested the new change and it's working as expected, @terrellt any thoughts before merging to master?

PR #18, opened 2015-03-16
Feature/cas

### LINKED ISSUE #4: CAS authentication
As an admin I want to be able to login using campus CAS so that I don't have to remember a new account and password.

wickr:
Fixes #4

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -8,11 +8,14 @@ Bundler.require(*Rails.groups)
 
 module ScholarsArchive
   class Application < Rails::Application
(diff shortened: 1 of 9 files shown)


config/config.yml
```diff
@@ -0,0 +1,4 @@
+rubycas:
```
terrellt: This should not be in the repo.

config/application.rb
```diff
@@ -8,11 +8,14 @@
 
 module ScholarsArchive
   class Application < Rails::Application
+    application_config = YAML.load_file(Rails.root.join('config/config.yml'))|| {}
```
terrellt: Can you set this to a constant like APP_CONFIG and use it in the devise initializer?

PR #69, opened 2015-05-06
Replaces content of deposit agreement with custom OSU agreement

### LINKED ISSUE #40: Create Deposit Agreement Text

straleyb:
Fixes #40. This simply gets all the text in place according to issue #40, however, layout could potentially use some work. We can decide on what to do with the layout of the text on that page anytime.


luisgreg99:
Looks good to me

PR #102, opened 2015-06-03
issue#89 add analytics to shared

zhang4952:


terrellt:
@zhang4952 Do you still need to do `rails g sufia:models:usagestats` and `rails g sufia:models:cached_stats`?

Issue #128, opened 2015-07-09
Make all solr fields stored in Sufia


terrellt:
This may not be feasible. It'd be so we can use atomic updates for metadata enrichment.

straleyb:
Make facetable, symbol, instead. and check that atomic updates in solr works.

zhang4952:
a/m/generic_file
a/c/cataloger_controller
solr atomic update

PR #149, opened 2015-07-15
updated gemfile to use hydra-editor to handle uris

### LINKED ISSUE #127: Enable Hydra-Editor to handle URIs
There's similar work in Oregon Digital 2, in @jechols' branch.

luisgreg99:
fixes #127


app/decorators/has_hint_option.rb
```diff
@@ -0,0 +1,15 @@
+class HasHintOption < SimpleDelegator
```
straleyb: This file is to tell people using the ingest form, which fields are CV's. @terrellt Do you think we should keep this in or wait until later to implement this feature?

app/models/generic_file.rb
```diff
@@ -3,17 +3,17 @@ class GenericFile < ActiveFedora::Base
   include Sufia::GenericFile
 
   property :spatial, predicate: ::RDF::DC.spatial do |index|
-    index.as :stored_searchable, :facetable
+    index.as :stored_searchable, :symbol
```
straleyb: I thought this was part of #128?

config/initializers/hydra_editor.rb
```diff
@@ -0,0 +1,5 @@
+HydraEditor.models = ["GenericFile"]
+HydraEditor::Fields::Generator.factory = ScholarsArchive::Fields::InputFactory.new(
+  HydraEditor::Fields::Factory,
+  DecoratorList.new(HasHintOption, HasURIInputType)
```
straleyb: If we decide to not use the HasHintOption quite yet, we can remove it from here as well.

config/routes.rb
```diff
@@ -10,9 +10,10 @@
     get "/users/sign_out", :to => "sessions#destroy"
   end
 
+ 
   # This must be the very last route in the file because it has a catch-all route for 404 errors.
   # This behavior seems to show up only in production mode.
   root to: 'homepage#index'
-
+  mount HydraEditor::Engine => '/'
```
straleyb: @luisgreg99 Can you explain why you needed this?

app/decorators/has_hint_option.rb
terrellt: Probably later.

app/forms/file_edit_form.rb
```diff
@@ -1,7 +1,16 @@
 # app/forms/file_edit_form.rb
 class FileEditForm < FilePresenter
   include HydraEditor::Form
+  delegate :validators, :to => :model
   include HydraEditor::Form::Permissions
-
+  self.model_class = GenericFile
   self.required_fields = [:title, :creator, :tag, :rights]
+
+  def property_hint(property)
```
straleyb: Since we dont need hints quite yet, we could take this method out.

app/models/generic_file.rb
luisgreg99: Yeah, you are right, I must have added it to see how the index would respond, but I forgot to change it back

straleyb: :+1:

config/routes.rb
luisgreg99: Sure, I was reading the readme file at https://github.com/jechols/hydra-editor/blob/master/README.md and it suggested to mount the routes in config/routes.rb. I think we might not need to mount it if we decide not to use the default paths.

straleyb: Ah sweet. I didn't see that. :+1: Looks good.

PR #151, opened 2015-07-16
#128: All fields are stored in Hydra

zhang4952:


app/schemas/scholars_archive_schema.rb
```diff
@@ -1,5 +1,6 @@
 class ScholarsArchiveSchema < ActiveTriples::Schema
 
+  property :subject, predicate: ::RDF::DC.subject
```
straleyb: We can probably remove this. Its one of the fields inherited from sufia.

spec/models/generic_file_spec.rb
```diff
@@ -1,7 +1,35 @@
 require 'rails_helper'
 
-RSpec.describe GenericFile do
-  it "should be able to persist" do
-    expect{GenericFile.new.save}.not_to raise_error
+RSpec.describe GenericFile do
```
straleyb: I think this test should be placed in enricher_spec instead of generic_file_spec.

straleyb:
Some of the objects in this PR need some tests as well. AltLabelEnhancement, LabelEnhancement, SolrProperty, and EnrichedSolrDocument.

spec/models/generic_file_spec.rb
zhang4952: It should be added back, but why enricher is a better place?

straleyb: Because the object that you are testing is not the GenericFile but the Enricher. If I were to look for the test that describes how the enricher works, I would expect to find it in the enricher_spec.rb, not the generic_file_spec.rb

spec/services/enricher_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end   
+
+    it "should return enriched solr field" do
+      enricher = Enricher.new(id)
+      enricher.enrich!
+      expect(ActiveFedora::SolrService.query("id:#{RSolr.solr_escape(id)}").first["subject_preferred_label_ssim"]).to eq ["Test"]
+    end
+  end
+
+  it "should be able to persist" do
```
straleyb: This test should still be in the generic_file_spec.rb Since you are testing the Generic File's capability of persisting.

PR #152, opened 2015-07-17
Feature/set up marmotta

straleyb:
This sets up Marmotta for scholars archive. To download the proper version of jetty 'marmotta-jetty' run 
```
rm -rf /jetty
rake jetty:clean
rake sufia:jetty:config
```
Then bundle install. This also utilizes the marmotta gem for ease of use.


straleyb:
@zhang4952 If you wanna check this out and try this branch out locally, go for it. You should be able to go to localhost:8983 and you should see a marmotta link with solr and fedora.

Gemfile
```diff
@@ -36,6 +35,8 @@ gem 'passenger'
 # MySQL
 gem 'mysql2'
 
+gem 'linkeddata', '~> 1.1.11.1', :github => 'ruby-rdf/linkeddata'
```
terrellt: Why do you need the latest master branch?

terrellt:
You may want https://github.com/OregonDigital/oregondigital_2/blob/master/spec/models/triple_powered_resource_spec.rb#L16-L32 to make sure this stuff works.

Gemfile
straleyb: Apparently there was a dependency on the version 1.1.11.1, Circle kept on complaining that it couldn't download this version from rubygems, so in order to make circle pass I had to explicitly grab this version from github. Is there an alternate way I could accomplish this?

straleyb:
:+1:

terrellt:
@straleyb Removed that line and did "bundle update linkeddata" and it seems to clear up.

config/settings.yml
```diff
@@ -0,0 +1,2 @@
+marmotta:
+  url: 'http://localhost:8983/marmotta'
```
terrellt: Should probably get this to work. Does Sufia already have a YML configuration file? Maybe install RailsConfig?

straleyb: I decided to install rails_config. Seems like it could be a useful tool for later.

straleyb:
@terrellt is this branch ready to be merged?

terrellt:
@straleyb You may have to delete the code besides repository in TriplePoweredResource. The label finder stuff isn't in SA.

PR #154, opened 2015-07-20
Feature/enable rdf

straleyb:
Adds RDF::URI Support but does not grab labels yet.


app/views/batch/_metadata.html.erb
```diff
@@ -0,0 +1,50 @@
+<h2>Individual Titles  </h2>
+<div id="" class="well">
+  <div class="row">
+    <div class="col-sm-8">
+      <h3>Applies to individual files uploaded</h3>
+      <%= content_tag :p, t('sufia.batch.help.title'), class: "help-block" %>
+      <% @batch.generic_files { |a,b| a.label.downcase <=> b.label.downcase }.each_with_index  do |gen_f, index| %>
```
straleyb: This is where that pesky line of code that calls #sort would be

@batch.generic_files.sort was the original code. Would overriding this file be a good way to fix this? Or should we send another PR  to Sufia to fix that. 

It seems like ActiveFedora does not want #sort being used on relations. 

https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/relation/delegation.rb#L10-L14

terrellt: #sort should work, just not #sort!

straleyb: https://github.com/projecthydra/sufia/blob/v6.0.0/app/views/batch/_metadata.html.erb#L7

The version of sufia we are using uses #sort!. Time to upgrade sufia?

terrellt: :+1:

PR #156, opened 2015-07-21
Feature/uri labels See #157

### LINKED ISSUE #131: The ingest form should have a button to switch between URIs and the labels they represent

straleyb:
Adds switchy button and enables labels to be fetched and stored in marmotta. Also changed up how the properties were defined for GenericFiles and other models. Fixes #131


straleyb:
@terrellt @zhang4952 @luisgreg99 This PR is ready for review.

app/presenters/file_presenter.rb
```diff
@@ -1,9 +1,4 @@
 # app/presenters/file_presenter.rb
 class FilePresenter < Sufia::GenericFilePresenter
-  self.terms = [:dc_type, :title, :creator, :contributor, :description,
-                :tag, :rights, :publisher, :date, :subject, :language,
-                :spatial, :temporal, :abstract, :tableOfContents, :format, 
-		:bibliographicCitation, :provenance, :isReferencedBy, :relation,
-		:rights, :title, :isCitedBy, :isIdenticalTo, :isPartOf,
-		:isVersionOf, :doi, :hdl, :orcid, :rid]
+  self.terms = ScholarsArchiveSchema.properties.map(&:name)
```
terrellt: Is this right? I thought it was self.terms +=, since GenericFile had some terms.

straleyb: Generic file contains the application of the ScholarsArchiveSchema 
```
apply_schema ScholarsArchiveSchema,
    ActiveFedora::SchemaIndexingStrategy.new(
      ActiveFedora::Indexers::GlobalIndexer.new([:stored_searchable, :symbol])
    ) 
```
Which contains properties. But I don't know if there are considered "terms" anymore.

terrellt: It also has the properties that are inherited from Sufia. I would expect that `terms =` would leave a bunch of missing stuff from your form (like title, for instance)

straleyb: Ahhh yeah they were not in the edit form, only the ingest form. I'll make the change and check that they are in the edit form.

straleyb: :+1:  They are back now.

terrellt:
@straleyb I can't get the labels to show up locally.

terrellt:
@straleyb Ah, so TriplePoweredResources aren't configured for Sufia's existing properties. Also, keyword doesn't save.

straleyb:
@terrellt Did you try killing the server and bringing it back? You may also have to do rm -f /jetty && rake jetty:clean && rake sufia:jetty:config if you still have the older version of jetty lying around

straleyb:
What do you mean TriplePoweredResources arent configured? Are you talking about in ScholarsArchiveSchema.rb?

terrellt:
> What do you mean TriplePoweredResources arent configured? Are you talking about in ScholarsArchiveSchema.rb?

So the fields that get added from the inheritance of Sufia (Title, Contributor, etc) don't have a configured class name, so those fields won't do label fetching. Try and put a URI in contributor.

straleyb:
Ah yeah it doesn't fetch the labels. I didn't even realize that. And it seems to have some trouble saving when you first create the GenericFile. If I create one, go edit it, then save it, it seems like the info saves. Just not the initial save.

terrellt:
That's because resque isn't running.

terrellt:
(And I can't seem to GET resque to run locally for some reason)

straleyb:
Hmm. Should I make a ticket for getting the resque working? And I'm going to add the sufia properties to the ScholarsArchiveSchema and see if that gets the labels working. But I have a feeling it might add 2 instances of those fields into the form.

terrellt:
It will.

straleyb:
So instead of using += should we use ~=?

terrellt:
You could move the property override off the schema and onto the object itself, but it's quite the cludge.

straleyb:
I'd like to stay away from that if possible, but if there is no other way, we can move the properties onto the object itself.

terrellt:
By property override I meant this: https://github.com/osulp/Scholars-Archive/pull/156/files#diff-41d4da347e7984b9622127ea10d49450R2

straleyb:
Woah hold on. I think you just blew my mind. Does FilePresenter's interface include #property? or even more specific Sufia::GenericFilePresenter.

terrellt:
No, but GenericFile's does.

PR #157, opened 2015-07-22
Feature/generic file cleanup

### LINKED ISSUE #131: The ingest form should have a button to switch between URIs and the labels they represent

straleyb:
This should replace #156 as it fixes keyword and uses the schema. Fixes #131.


app/schemas/scholars_archive_schema.rb
```diff
@@ -0,0 +1,45 @@
...
+  property :orcid, predicate: ::RDF::Vocab::Identifiers.orcid
+  property :rid, predicate: ::RDF::Vocab::Identifiers.rid
+ 
+
+  def self.sufia_default
+    Sufia::GenericFilePresenter.terms
+  end
+
+  def self.presenter_properties
+    properties.map(&:name).delete_if { |x| sufia_default.include?(x) }
```
terrellt: Should be able to just do
```ruby
  properties.map(&:name) - sufia_default
```

terrellt: I should note, it's a little concerning if we're adding terms that already exist in Sufia. Maybe we should just remove the duplicates from our schema?

terrellt: Oh, but you probably duplicated them on purpose so you could add the TriplePoweredResource.

straleyb: @terrellt Yeah I added the duplicates to make sure they were able to be represented as URI's. I'm going to test really quickly removing title and using a URI and seeing if it works properly.

PR #171, opened 2015-08-18
Fix conflicting schemas.

terrellt:


terrellt:
When the schemas conflict the forms for Sufia stop persisting.

PR #173, opened 2015-08-18
added default publisher and language in edit form

### LINKED ISSUE #136: Ingest Form - Prepopulate publisher
Prepopulate the Publisher field to Oregon State University for all items

### LINKED ISSUE #139: Ingest Form - prepopulate language
PrePopulate the Language field with English using an appropriate controlled vocabulary

luisgreg99:
fixes #136 
fixes #139


app/controllers/batch_controller.rb
```diff
@@ -3,4 +3,12 @@ class BatchController < ApplicationController
   include Sufia::BatchControllerBehavior
 
   self.edit_form_class = BatchEditForm
+
+  def edit
+    @batch = Batch.find_or_create(params[:id])
```
straleyb: Could this be refactored into a private method that gets the batch?

straleyb:
@luisgreg99 I'd like to see a test to show that the information appears where you would expect it to.

spec/features/edit_generic_files_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  before do
+    login_as(user)
+  end
+
+  it "should display a default publisher" do
+    visit sufia.edit_generic_file_path(file) 
+    expect(find_field('Publisher').value).to eq publisher
+    click_button 'Update Generic file'
```
terrellt: Why actually persist these?

spec/features/edit_generic_files_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
+
+describe "edit form and set default values", type: :feature do
```
terrellt: I don't think this is necessary.

app/controllers/batch_controller.rb
```diff
@@ -1,6 +1,17 @@
 # app/controllers/batch_controller.rb
 class BatchController < ApplicationController
   include Sufia::BatchControllerBehavior
-
   self.edit_form_class = BatchEditForm
+
+  def edit
+    super
+    defaults_object.set_default_values
```
terrellt: There's an API for this.

```ruby
def edit
  super
  @form.attributes = default_values
end

private

def default_values
    {
      :publisher => I18n.t('form_defaults.publisher'),
      :language => I18n.t('form_defaults.language_uri')
    }
end
```

app/services/default_values_object.rb
```diff
@@ -0,0 +1,21 @@
+class DefaultValuesObject
```
terrellt: Okay, these are just my opinions, but here goes.

This is a service object which takes a single parameter, which there's no API to access nor would you want to, for instantiation. There are no methods that reference it, so you're not simplifying a method signature or allowing for injected dependencies. Thus, this class is effectively equivalent to

```ruby
class DefaultValuesObject
  def set_default_values(input_object)
    default_values.each_pair do |key, value|
      input_object[key] = value if input_object[key].first.empty?
    end
    input_object
  end
end
```

Now the API is default_values_object.set_default_values(@edit_form)

Since you've rewritten it this way, you can see the classic anti-pattern of feature envy (https://sourcemaking.com/refactoring/feature-envy). This class basically becomes a hash holder for things that are always called on `input_object`

The other problem is this object has one method and no interface except that one method, which takes the arguments it needs for that method - that's a function, which is effectively namespaced by a constant (DefaultValuesObject). It would be just as effective if this was `DefaultValuesObject.set_default_values(input_object)`, which is just a function - and ruby's OO.

straleyb: Ah I see. So instead what we could have done is look at the methods on the form to see if there are setters for its attributes. I think I'm going to do more research on feature-envy so next time we come across this, It'll be easier to spot and mitigate.

terrellt: Even if there wasn't a method there for it already, defining something on the object you're acting on would be the better way.

straleyb: Yeah, the more I'm reading about it, Im seeing that whatever operations you are doing on an objects data (in this case, on the data utilized by the forms), you should define those methods where the data needs to be. In this case, would it also be useful to define those defaults in the form object also, instead of in the controller?

luisgreg99: So after reading the post about Feature Envy, to avoid the anti-pattern, we could move back our `set_default_values` method in the `BatchController`, which is so far the only place where the `@form` variable gets modified to account for default values. If I understand correctly, from the refactoring post, the rule is to place or define the new feature where the data gets changed right?

app/forms/batch_edit_form.rb
```diff
@@ -1,3 +1,4 @@
 # app/forms/batch_edit_form.rb
 class BatchEditForm < FileEditForm
+  delegate :attributes, :to => :model
```
terrellt: No longer needed.

app/controllers/batch_controller.rb
```diff
@@ -1,6 +1,21 @@
 # app/controllers/batch_controller.rb
 class BatchController < ApplicationController
   include Sufia::BatchControllerBehavior
-
   self.edit_form_class = BatchEditForm
+
+  protected
```
terrellt: Newline after protected please.

luisgreg99: done

PR #175, opened 2015-08-19
Feature/remove elements

zhang4952:


app/presenters/file_presenter.rb
```diff
@@ -1,4 +1,5 @@
 # app/presenters/file_presenter.rb
 class FilePresenter < Sufia::GenericFilePresenter
   self.terms += ScholarsArchiveSchema.properties.map(&:name)
+  self.terms -= [:subject]
```
straleyb: I feel like you should be able to do:
```
self.terms += ScholarsArchiveSchema.properties.map(&:name) - [:subject]
```

PR #176, opened 2015-08-19
Added add and remove buttons where needed

### LINKED ISSUE #159: Need Remove buttons on ingest form fields
When one adds a new field in the ingest form the new field does not have a remove button, but it should

straleyb:
Fixes #159

This adds a remove button to the last element of a group of fields under a certain metadata field. 
    1)When the form is first visited, the last field will only have the add button
    2)When removing a field not at the bottom of the list, it keeps both buttons on the last field
    3)When removing the field at the bottom, it adds the add and remove buttons to the field above it.
    4)When adding a field, the add button will move to the newly created field with a remove button.

### DIFF
--- a/app/assets/javascripts/hydra-editor/manage_repeating_fields.js
+++ b/app/assets/javascripts/hydra-editor/manage_repeating_fields.js
@@ -1,0 +1,156 @@
+// This widget manages the adding and removing of repeating fields.
+// There are a lot of assumptions about the structure of the classes and elements.
+// These assumptions are reflected in the MultiValueInput class.
+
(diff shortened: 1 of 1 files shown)


terrellt:
You should demo this with @vantuyls

straleyb:
Good idea! :+1: @vantuyls Is there some time you could meet for a few minutes today to show you how this is working? Or we could meet tomorrow after or before standup if you wanna.

terrellt:
@vantuyls Did this look good?

vantuyls:
yep, looks great.

PR #181, opened 2015-08-21
Nested authors

terrellt:


app/models/generic_file.rb
```diff
@@ -1,7 +1,7 @@
 # app/models/generic_file.rb
 class GenericFile < ActiveFedora::Base
   def self.property(name, options)
-    super(name, options.merge(:class_name => TriplePoweredResource)) do |index|
+    super(name, {:class_name => TriplePoweredResource}.merge(options)) do |index|
```
straleyb: Why switch this around?

terrellt: So that if the property has a defined class name, this doesn't override it.

app/assets/javascripts/nested_fields.js.coffee
```diff
@@ -0,0 +1,30 @@
+jQuery ->
+  $(".nested-field").manage_fields()
+  $(".nested-field").on("managed_field:remove", (event, removed) ->
```
straleyb: I understand where the event comes from, but where does removed come from?

terrellt: It's the argument passed when the event is fired.

PR #198, opened 2015-09-01
Feature/date stuff

### LINKED ISSUE #138: Date fields on the ingest form with a drop-down list of date types
Add Other Date field to form - allow dropdown list of date types (from DataCite)

luisgreg99:
Closes #138


straleyb:
@wickr Greg and I worked on this PR together. So we would love to have you review it.

app/assets/javascripts/add_date_field.js.coffee
```diff
@@ -0,0 +1,32 @@
...
+    event.preventDefault()
+    
+    #Variable assignment and necessary string manipulation
+    notice = '<div class="alert alert-info alert-dismissible" role="alert"><button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button> That date field already exists </div>'
+    type = $('#new_date_type').val()
+    type_string = type_manipulation(type)
+    date_created = $(".form-group.generic_file_date_created")
+    html = html_manipulation(date_created.clone(), type, type_string)
+
+    html.find("ul.listing li:not(:last-child)").remove()
```
wickr: Is this specific enough to only affect the date field stuff? I think any change to the form could break this.

app/forms/file_edit_form.rb
```diff
@@ -6,6 +6,11 @@ class FileEditForm < FilePresenter
   include HydraEditor::Form::Permissions
   self.model_class = GenericFile
   self.required_fields = [:title, :rights, :keyword]
+  self.terms -= [:accepted, :available, :copyrighted,:created, :issued,:submitted,:modified, :valid_date]
```
wickr: Consistent spacing here please.

spec/features/edit_batch_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  before do
+    login_as user, :scope => :user
+    visit "/batches/#{file.batch.id}/edit"
+    page.find('#show_addl_descriptions').click
+  end
+
+  context "when ingesting a file", js:true do
+    it "should display proper date fields" do
```
wickr: Add something on the end of the string to the effect of "...when the Add Date button is clicked". Could also separate the tests out so it checks that the new date fields don't show by default, and then does another check to make sure they do.

app/assets/javascripts/add_date_field.js.coffee
straleyb: I have reason to believe this is specific enough to only target date field stuff. This picks out the listing from the html we parsed out, which is a copy of the 'generic_file_date_created' section of the form, of which there is only one.

spec/features/edit_batch_spec.rb
straleyb: Yeah we could split this up into two tests. It would be something to the effect of 
```
it 'should only display the default date field' do
....
end
```
and 
```
context 'after a field is selected and the button is clicked' do
   it 'should display the selected field' do
        .......
   end
end

File: app/presenters/file_presenter.rb [hunk 4]
```diff
@@ -1,7 +1,7 @@
 # app/presenters/file_presenter.rb
 class FilePresenter < Sufia::GenericFilePresenter
   self.terms += ScholarsArchiveSchema.properties.map(&:name)
-  self.terms -= [:subject, :keyword, :creator]
+  self.terms -= [:subject, :keyword, :creator, :accepted, :available, :copyrighted,:created, :issued,:submitted,:modified, :valid_date]
```
wickr: Consistent spacing here too.

app/assets/javascripts/add_date_field.js.coffee
luisgreg99: I updated the selector in the default date field so that it is more specific to the dates section

app/presenters/file_presenter.rb
luisgreg99: done

spec/features/edit_batch_spec.rb
luisgreg99: I added another test case for the default behavior when ingesting the file

straleyb:
@wickr Can you re-review this? Its been updated. Its the changes we made to the form so you can choose different types of dates from a drop down.

PR #200, opened 2015-10-07
enable statistics dashboard for admin

zhang4952:


straleyb:
Can you explain what this initializer and changing the solr config is supposed to do? Is this supposed to allow access to a specific webpage?

zhang4952:
config admin_stats initializer:
required to make statistics dashboard available for admin
solrconfig.xml:
add termComponent because it is used by statistics dashboard

straleyb:
Ok! If its a route thats normally unavailable to normal users but available to admins, Id like to see a test for that.

vantuyls:
Wait, is this done? It's not running right now? If this isn't completed we need to move it back to a different column...

straleyb:
When we merged it, it auto closed the ticket I think. Ill move it to in progress and change the verbage.

PR #235, opened 2015-11-12
Feature/finished embargo

### LINKED ISSUE #227: Add Hydra::AccessControls::Embargoable to necessary models
GenericFile

### LINKED ISSUE #228: Setup Embargo actor and actor manager
https://github.com/ualbertalib/HydraNorth/blob/49d109962a3ca6ac532254ccdf1b2df369dd6c93/app/actors/hydranorth/manages_embargoes_actor.rb

https://github.com/ualbertalib/HydraNorth/blob/49d109962a3ca6ac532254ccdf1b2df369dd6c93/app/actors/hydranorth/generic_file/actor.rb

### LINKED ISSUE #229: Build actor in generic files controller behavior
https://github.com/ualbertalib/HydraNorth/blob/49d109962a3ca6ac532254ccdf1b2df369dd6c93/app/controllers/concerns/hydranorth/files_controller_behavior.rb

### LINKED ISSUE #230: Build view to assign embargo on generic file
On the edit page (when generating a new asset, it creates it initially then takes you to the edit page)

### LINKED ISSUE #232: Change styling and wording of visibility options
currently, visibility settings include one level titled: Oregon State University Libraries & Press, this should be changed to read: Oregon State University. 

Also, can the background of that visibility level be set to OSU orange?

straleyb:
This is all the necessary code to allow for embargoes in Scholars Archive. The system will automatically detect when an embargo has run out and flag it as needing to be processed. Then one can run a rake task to process said expired embargoes. This also does some various formatting as well as adds a date picker in for use. 

Fixes #227 Fixes #228 Fixes #229 Fixes #230 Fixes #232


app/views/generic_files/_embargo_help.html.erb
```diff
@@ -0,0 +1,7 @@
+<p>Embargo description goes here...</p>
```
luisgreg99: @straleyb I think we don't need this form any more right? I had added this partial form for the help bubble. Since we agreed that the embargo option would go under visibility, we could move this form and merge it with https://github.com/projecthydra/sufia/blob/master/app/views/generic_files/_visibility.html.erb

spec/actors/generic_files/actor_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
```
luisgreg99: I think we can remove this file, it appears to be a duplicate of https://github.com/osulp/Scholars-Archive/blob/feature/FinishedEmbargo/spec/actors/scholars_archive/generic_files/actor_spec.rb

PR #259, opened 2016-01-08
Fixed language edits and tag

### LINKED ISSUE #195: Default publisher should be label (with URI) for OSU

### LINKED ISSUE #196: Prepopulating with URI for language should have switchy button

straleyb:
Fixes #195 Fixes #196


app/schemas/scholars_archive_schema.rb
```diff
@@ -1,6 +1,6 @@
 class ScholarsArchiveSchema < ActiveTriples::Schema
-  property :tag, predicate: ::RDF::DC11.subject
-  property :language, predicate: ::RDF::DC11.language
+
```
wickr: Looks like :language disappeared from the schema?

app/helpers/sufia_helper.rb
```diff
@@ -15,4 +15,19 @@ def link_to_fields(fields, field_string)
...
+    else
+      p = { search_field: 'advanced', fieldname => '"' + fieldvalue + '"' }
+    end
+    link_url = catalog_index_path(p)
+    if fieldvalue["preflabel"]
+      display = displayvalue.blank? ? fieldvalue["preflabel"] : displayvalue
+    else
+      display = displayvalue.blank? ? fieldvalue : displayvalue
+    end
+    display = displayvalue.blank? ? fieldvalue : displayvalue
```
mickeroo: Is this a duplicate of 28?  Seems like you're doing a test then overwriting the result.

straleyb: Are you talking about issue 28? Ill investigate.

wickr: @straleyb no, line 28. 28 and 30 look to be the same.

straleyb: Ahh yeah it looks to be so. That would override everything that was done in the if statement. Good catch! Ill make the adjustments.

app/helpers/sufia_helper.rb
```diff
@@ -15,4 +15,18 @@ def link_to_fields(fields, field_string)
     fields.map { |tag| link_to_facet(tag, field_string) }
   end
 
+  def link_to_field(fieldname, fieldvalue, displayvalue = nil)
```
wickr: Can you add a comment about why this needs to be here and what it does differently? Looks like it's only used for nested_authors?

app/views/records/show_fields/_language.html.erb
```diff
@@ -0,0 +1,3 @@
+<% record.language.each do |lang| %>
+  <span itemprop="Language"><%= link_to_facet(lang, Solrizer.solr_name("language", :facetable)) %></span><br />
```
wickr: I think `itemprop='Language'` should be lowercase.

app/helpers/sufia_helper.rb
straleyb: Yeah I think this was for an older way I was unpacking the rdf objects, trying to reuse the changes trey had made. But in the partials, im not using link to field, but link to facet. So I can change this back and it should be fine.

app/views/records/show_fields/_language.html.erb
straleyb: :+1:

PR #265, opened 2016-01-21
Add English as default label

### LINKED ISSUE #257: Label for English is coming through as Englisch (the german label)
it shouldn't do that

zhang4952:
fix #257

### DIFF
--- a/app/controllers/batch_controller.rb
+++ b/app/controllers/batch_controller.rb
@@ -14,8 +14,8 @@ class BatchController < ApplicationController
     {
       "creator" => [current_user.name],
       "title" => @batch.generic_files.map(&:label),
-      "publisher" => [I18n.t('form_defaults.publisher')],
-      "language" => [I18n.t('form_defaults.language_uri')]
+      "publisher" => [TriplePoweredResource.new(RDF::URI(I18n.t('form_defaults.publisher')))],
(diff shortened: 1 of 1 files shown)


app/controllers/batch_controller.rb
```diff
@@ -14,8 +14,8 @@ def default_values
     {
       "creator" => [current_user.name],
       "title" => @batch.generic_files.map(&:label),
-      "publisher" => [I18n.t('form_defaults.publisher')],
-      "language" => [I18n.t('form_defaults.language_uri')]
+      "publisher" => [TriplePoweredResource.new(RDF::URI(I18n.t('form_defaults.publisher')))],
+      "language" => [I18n.t(TriplePoweredResource.new(RDF::URI(I18n.t('form_defaults.language_uri'))), :default => 'English')]
```
straleyb: If you save this with English as the default, is it saving the string as the default or the URI?

wickr:
This does not solve #257. The problem is not that the default label of the form is the wrong 'English'. The problem is that when the labels are fetched from the English URI, which is already in the system as the default, when the RDF label is pulled for the value/facet, it picks the wrong one.

We shouldn't need a default URI and a default label, because if the URI is there the label from the form shouldn't be saved. This is a fetch and retrieval issue, not a form issue.

PR #283, opened 2016-02-11
Added ability to delete researchers

### LINKED ISSUE #160: Featured Researcher - need to be able to delete old ones
Need to be able to remove Featured Researchers from the "Other Featured Researchers" list

straleyb:
Fixes #160 This allows for the deletion of old featured researchers.


config/routes.rb
```diff
@@ -6,6 +6,8 @@
   get '/help/:page', :to => "help#page"
   get '/help', :to => "help#page", :page => "general"
 
+  get '/featured_researcher/:id', :to => "content_blocks#delete", :as => "delete_researcher"
```
wickr: You'll want a Delete verb instead of Get if you're going directly to #delete. What you have is usually the route for show.

An example from Sufia's routes is:
```
delete 'users/:user_id/depositors/:id' => 'depositors#destroy', as: 'user_depositor'
```
at https://github.com/projecthydra/sufia/blob/master/config/routes.rb#L47

wickr:
spec/controllers/content_blockes_controller_spec.rb   has an extra 'e' in it

PR #321, opened 2016-03-18
Using Preferred label on show page and in form

### LINKED ISSUE #302: Englisch to English

straleyb:
Fixes #302 Sadly, most of the time an array of multiple different labels gets returned on the show page and in the form. Now, instead of using the first option which may not be the english label, we are now using the uri and its preferred label after it is cast to a triple powered resource. This should guarantee the label we have to always be the preferred language which is english.

### DIFF
--- a/app/helpers/sufia_helper.rb
+++ b/app/helpers/sufia_helper.rb
@@ -5,7 +5,7 @@ module SufiaHelper
 
   def link_to_facet(field, field_string)
     if field['preflabel']
-      link_to(field['preflabel'].first, catalog_index_path(add_facet_params(field_string, field['id'])))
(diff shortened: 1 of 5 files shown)


app/helpers/sufia_helper.rb
```diff
@@ -5,7 +5,7 @@ module SufiaHelper
 
   def link_to_facet(field, field_string)
     if field['preflabel']
-      link_to(field['preflabel'].first, catalog_index_path(add_facet_params(field_string, field['id'])))
+      link_to(preffered_uri_label(field), catalog_index_path(add_facet_params(field_string, field['id'])))
```
revgum: nitpick; preferred is misspelled..

PR #326, opened 2016-03-18
adds a reasonable baseline for log rotation to the app

### LINKED ISSUE #325: Production/Staging log file management
Prevent the application log from growing indefinitely and filling the disk.

revgum:
fixes #325


wickr:
10.megabytes and 1.megabytes seems on the small side to me, I think those could rotate out a lot. Any thoughts @decimalator  ?

revgum:
decided that we'll use logrotate and manage all of the logs in the shared directly at once

PR #332, opened 2016-03-25
Feature/geo coordinates

### LINKED ISSUE #144: [Geo] Nested attributes for geographic point
Relates to #301 and #300. Some items will have geographic information that is not in Geonames (nor should it be?) - example: a latitude and longitude pair, or the bounding coordinates for an area

### LINKED ISSUE #250: [Geo] Offer labels for geographic points and bounding boxes
e.g. this point is ryan's house, this other point is steve's house

luisgreg99:
fixes #144 
fixes #250


app/controllers/generic_files_controller.rb
```diff
@@ -9,11 +9,29 @@ class GenericFilesController < ApplicationController
   self.edit_form_class = FileEditForm
 
   def update_metadata
+    params["generic_file"]["nested_geo_bbox_attributes"].each do |box, value|
+      if [value["label"], value["bbox_lat_north"], value["bbox_lon_west"], value["bbox_lat_south"], value["bbox_lon_east"]].none? { |f| f.empty? }
```
revgum: I think you would want _f.blank?_ here to handle values that could be empty or nil, in case one or more of the fields are not posted in the form.

app/assets/javascripts/add_geo_field.js.coffee
```diff
@@ -0,0 +1,9 @@
+jQuery ->
+  $(".geo_bbox").hide() if $("#generic-file-location-form .geo_bbox ul.listing li > .form-group").hasClass("generic_file_nested_geo_bbox_id") is false
```
revgum: You want the same function for .geo_point so that it is hidden by default as well.

`$(".geo_point").hide() if $("#generic-file-location-form .geo_point ul.listing li > .form-group").hasClass("generic_file_nested_geo_point_id") is false`

app/views/records/edit_fields/_nested_geo_location_form.html.erb
```diff
@@ -0,0 +1,43 @@
...
+        <h4>Bounding box</h4>
+        <%= f.simple_fields_for :nested_geo_bbox do |a| %>
+          <li class="field-wrapper">
+            <% unless a.object.new_record? %>
+              <%= a.input :id, :as => :hidden %>
+            <% end %>
+
+            <%= a.input :label, :wrapper_html => {:class => "col-xs-12"} %>
+            <%= a.input :bbox, :wrapper_html => {:class => "col-xs-12 bbox_field"} %>
+            <%= a.input :bbox_lat_north, :wrapper_html => {:class => "col-xs-6"} %>
```
revgum: bbox_lat_north, bbox_lat_west, bbox_lat_south, and bbox_lat_east could use nice labels for the form, you can override the form label like so;

`<%= a.input :bbox_lat_north, :label => "Latitude North", :wrapper_html => {:class => "col-xs-6"} %>`

app/controllers/generic_files_controller.rb
```diff
@@ -9,11 +9,29 @@ class GenericFilesController < ApplicationController
...
+    end
     file_attributes = edit_form_class.model_attributes(params[:generic_file])
     updated_attributes = AttributeURIConverter.new(file_attributes).convert_attributes
     actor.update_metadata(updated_attributes, params[:visibility])
   end
 
+  # routed to /files/:id/edit
+  def edit
+    @generic_file.nested_geo_bbox.each do |box|
+      box_array = box.bbox.to_a.first.split(",").map { |s| s.tr('[]" ','').to_s }
```
revgum: This line seems like it is worthy of a slight refactor and comment so that our future self will understand what the intent it.

`# bbox is stored as a string array of lat/long string arrays like: '["121.1", "121.2", "44.1", "44.2"]', however only one array of lat/long array is stored, so the first will need to be converted to simple array of strings like: ["121.1","121.2","44.1","44.2"]
box_array = box.bbox.to_a.first.tr('[]" ','').split(',')
`

PR #387, opened 2016-04-28
Fix/rspec

revgum:


revgum:
will open as a new PR when work is ready

PR #394, opened 2016-05-04
Feature/date dropdown and property

### LINKED ISSUE #369: [S6 -> S7 port] Add date properties to metadata fields
Relates to #365. We need to make sure that the date properties are added to the system. We don't want to add them to the form as it would normally be generated. Instead what we want to do is add one date field and have a selector with other date type fields that we could add dynamically depending on what is needed. So for this ticket we need to make sure that the date properties exist in the system as well as dont show up in the ingest form.

- [ ] Date properties added to system
- [ ] Date properties dont show up on ingest or edit forms.

### LINKED ISSUE #370: [S6 -> S7 port] Create date select box and make it work for existing date fields
This should be a select on the ingest forms. This is to select different types of dates depending on what is needed for the work. There is some javascript involved that has been written for Scholars Archive, but the way this was written could be changed.

straleyb:
Fixes #369 Fixes #370 

This adds in the functionality of the date dropdown and the dynamic dates. The dates show up on the show page for the item. A user can edit and re-save and it updates the information.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,14 +36,6 @@ gem 'mysql2', '~> 0.3.18'
 
 gem 'resque'
 
-group :development, :test do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-  gem 'byebug'
-  gem 'pry-rails'
-  gem 'pry'
-  gem 'pry-byebug'
-end
-
 gem 'capistrano', '~> 2.0'
 
 group :development do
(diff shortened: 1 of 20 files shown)


revgum:
I think our config/locales/sufia.en.yml could benefit from updates to make the simple form field labels consistent on the edit and the show forms.. "Date Created" instead of "Date created" on the edit form, and then when viewing a generic work, "Date Published" instead of "Published" (and etc for other fields).. check Sufia projects example? https://github.com/projecthydra/sufia/blob/master/config/locales/sufia.en.yml

PR #441, opened 2016-08-03
Adds roles and toggle feature/unfeature

### LINKED ISSUE #377: [S6 -> S7 port] Feature/Unfeature item from item record page
For user experience purposes, we need to be able to feature/unfeature items from the item record page. The current workflow is you can feature a researcher from the item record page, but cannot unfeature unless you are on the root page of the website.

straleyb:
This adds role management to Scholars archive and returns the functionality of being able to feature and unfeature an item from the item record page. Fixes #377

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -8,6 +8,10 @@ class Ability
 
   # Define any customized permissions here.
   def custom_permissions
+    if current_user.admin?
+      can [:create, :show, :add_user, :remove_user, :index, :edit, :update, :destroy], Role
+    end
+
     # Limits deleting objects to a the admin user
     #
     # if current_user.admin?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,7 @@
 class User < ActiveRecord::Base
   # Connects this user object to Hydra behaviors.
   include Hydra::User
+
   # Connects this user object to Role-management behaviors.
   include Hydra::RoleManagement::UserRoles
   # Connects this user object to Curation Concerns behaviors.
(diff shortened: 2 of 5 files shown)


app/models/user.rb
```diff
@@ -3,6 +3,10 @@ class User < ActiveRecord::Base
   include Hydra::User
   # Connects this user object to Role-management behaviors.
   include Hydra::RoleManagement::UserRoles
+
+
+  # Connects this user object to Role-management behaviors.
+  include Hydra::RoleManagement::UserRoles
```
wickr: Looks like a duplicate of line #5

app/models/ability.rb
```diff
@@ -8,6 +8,10 @@ class Ability
 
   # Define any customized permissions here.
   def custom_permissions
+    if current_user.admin?
+      can [:create, :show, :add_user, :remove_user, :index, :edit, :update, :destroy], Role
```
wickr: Should this check against the user group and not just their abilities?  There's this in CC too: https://github.com/projecthydra/curation_concerns/blob/master/app/models/concerns/curation_concerns/ability.rb#L40

wickr:
app/assets/javascripts/sufia/featured_works.js  is an emtpy file?

app/models/ability.rb
straleyb: In sufia 7 user groups dont exist anymore. Instead what you do is add a user to a role and then you can check whether a role contains a user.

wickr: Ok, but the wiki page and gem haven't been updated since 2015, so not sure what that has to do with Sufia 7. But if that works now that I guess that's fine. There's this too: https://github.com/projecthydra/sufia/issues/2032

PR #447, opened 2016-08-12
Feature/delete researcher role management

### LINKED ISSUE #376: [S6 -> S7 port] Delete featured researcher
This was a fairly quick fix in ScholarsArchive on sufia 6. Basic crud actions for a researcher. We need to be able to delete featured researchers so the list of previously featured researchers doesnt get too long.

straleyb:
Fixes #376


app/controllers/delete_featured_researchers_controller.rb
```diff
@@ -0,0 +1,22 @@
+class DeleteFeaturedResearchersController < ApplicationController
```
wickr: Would this be better off extending the ContentBlock model/controller from Sufia than it's own controller? Either way, I don't think 'Delete' should be part of the controller name.

PR #469, opened 2016-09-21
configured sidekiq server, and deploy cleanup

### LINKED ISSUE #467: sidekiq web ui admin link?
Introduce a link to /sidekiq for users who are administrators, probably in the users link dropdown in the upper right of the screen?

revgum:
fixes #467 

changes sidekiq address to something more generic for if/when sidekiq is replaced

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -1,0 +1,36 @@
+<ul id="user_utility_links" class="nav navbar-nav navbar-right">
+  <% if user_signed_in? %>
+      <% if current_user.admin? %>
+          <li>
+            <%= link_to "Queues Monitor", root_url + "queues-monitor" %>
+          </li>
+      <% end %>
+      <li>
+        <%= link_to sufia.notifications_path do %>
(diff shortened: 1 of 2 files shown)


app/views/_user_util_links.html.erb
```diff
@@ -0,0 +1,36 @@
+<ul id="user_utility_links" class="nav navbar-nav navbar-right">
+  <% if user_signed_in? %>
+      <% if current_user.admin? %>
```
straleyb: Could bring this into one line with 
```
if user_signed_in? && current_user.admin?
```

straleyb:
Looks good besides that one comment. 👍

app/views/_user_util_links.html.erb
revgum: no, actually these are two separate functionalities.. only admin's should see the link, while all signed in users should see the other stuff as well

straleyb: Ahh ok yeah I see. Cool dude. Merging.

PR #471, opened 2016-09-22
Add ability to delete featured researchers

straleyb:


spec/controllers/content_blocks_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        r.save
+        u
+      end
+
+      before do
+        sign_in(user) if user
+      end
+
+      it "should delete a featured_researcher" do
+
```
wickr: Can you remove the extra linebreaks?

app/controllers/content_blocks_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def index
+  end
+
+  def destroy
+    @researcher = ContentBlock.find(params[:id])
+    if @researcher.destroy
+      flash[:success] = "Sucessfully deleted."
+    else
+      flash[:error] = "Error in deleting entry."
```
wickr: I think it'd be good to be a little more specific with the error, maybe "Error while deleting Featured Researcher."

circle.yml
```diff
@@ -9,7 +9,6 @@ dependencies:
   pre:
     - gem install fcrepo_wrapper
     - gem install solr_wrapper
-    - npm install phantomjs@1.9.8
```
revgum: Think we want this back in the mix, right?

app/controllers/content_blocks_controller.rb
```diff
@@ -0,0 +1,53 @@
+class ContentBlocksController < ApplicationController
```
revgum: Definitely extend Sufia's controller, that'll make the test and updates easier

circle.yml
straleyb: Im not too sure. The tests seemed to pass without it. Im curious to see if they will keep passing. Ill add it back in, but if we can eventually take it out, that would be awesome.

vantuyls:
thumbs up.

PR #486, opened 2017-02-15
Add default metadata as module; and two work types: default_work, etd

zhang4952:
This branch is for DSpace migration and should be merged to Hyrax

### DIFF
--- a/app/models/concerns/scholars_archive/default_metadata.rb
+++ b/app/models/concerns/scholars_archive/default_metadata.rb
@@ -4,8 +4,57 @@ module ScholarsArchive
 
     #reusable metadata fields for DSpace migration 
     included do
-      
+	  property :orcid, predicate: ::RDF::URI.new("http://id.loc.gov/vocabulary/identifiers/orcid") do |index|
+	    index.as :stored_searchable
+	  end
 
+	  property :date_issued, predicate: ::RDF::URI.new("http://purl.org/dc/terms/issued") do |index|
+	    index.as :stored_searchable, :facetable
+	  end
+
+	  property :date_embargo, predicate: ::RDF::URI.new("http://purl.org/dc/terms/description") do |index|
+	    index.as :stored_searchable
+	  end
+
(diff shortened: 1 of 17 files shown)


app/models/concerns/scholars_archive/default_metadata.rb
```diff
@@ -4,8 +4,57 @@ module DefaultMetadata
...
+
+	  property :date_available, predicate: ::RDF::URI.new("http://purl.org/dc/terms/description") do |index|
+	    index.as :stored_searchable
+	  end
+
+	  property :date_updated, predicate: ::RDF::URI.new("http://purl.org/dc/terms/date") do |index|
+	    index.as :stored_searchable
+	  end
+
+	  property :citation, predicate: ::RDF::URI.new("http://purl.org/dc/terms/identifier") do |index|
```
wickr: This isn't right for Citation.

app/models/concerns/scholars_archive/default_metadata.rb
```diff
@@ -4,8 +4,57 @@ module DefaultMetadata
...
+	    index.as :stored_searchable
+	  end
+
+	  property :isbn, predicate: ::RDF::URI.new("http://purl.org/dc/terms/identifier") do |index|
+	    index.as :stored_searchable
+	  end
+
+	  property :identifier_uri, predicate: ::RDF::URI.new("http://purl.org/dc/terms/identifier") do |index|
```
wickr: We can't have multiple properties with the same predicate.

app/models/concerns/scholars_archive/default_metadata.rb
zhang4952: citation is a kind of identifier now, as dc.identifier.citation

app/models/concerns/scholars_archive/default_metadata.rb
zhang4952: how about identifieruri, or uri_identifier?

PR #523, opened 2017-04-18
added additional date fields and updated the form to include the date…

### LINKED ISSUE #138: Date fields on the ingest form with a drop-down list of date types

### LINKED ISSUE #396: [Dates] Identify which date types are repeatable versus nonrepeatable
And then make it so.

### LINKED ISSUE #514: UI Mockup for Date and Geo Picker
Need to create UI mockups for the date picker and the geo picker for ingest forms

luisgreg99:
… type select box

fixes #138 
fixes #396 
fixes #514 

New/Edit forms animation

date-range-iso-8601

New/Edit forms screenshot

image

Show page 

image

Search results page

image

TODO
- [x] Expose dates in show pages
- [x] Expose date fields in search results 
- [x] Verify proper language in date field descriptions
- [x] Provide proper translations
- [x] Add DateTimePicker plugin


app/forms/hyrax/default_work_form.rb
```diff
@@ -4,5 +4,29 @@ module Hyrax
...
+      date_terms.each do |date_term|
+        permitted << {
+          date_term => []
+        }
+      end
+    end
+
+    def date_terms
+        [
+          :date_created,
```
revgum: Is `date_created` supposed to be a single value field? 

@vantuyls might help to verify the multi/single valuedness of each of these fields.

app/views/hyrax/base/_form_dates.html.erb
```diff
@@ -0,0 +1,27 @@
+<div id="date_wrapper">
+  <div class="row"> 
+    <legend>Dates</legend>
+    <div class="warning-anchor"></div> 
+    <div class="col-sm-6">
+      <select name="new_date_type" id="new_date_type" class="form-control">
+        <option value="Select a date type">Select a date type</option>
+        <option value="accepted">Accepted</option>
```
revgum: we should probably try to make these options supplied by the form, so that the value and text are a bit more dynamic. I'm thinking the first option "Select a date type" is good, but then mapping the date_terms..maybe something along these lines:

`f.object.date_terms.map { |t| "<option value='#{t.to_s}'>#{t.humanize}</option>" }`

I'm not 100% sure on the syntax.. but the goal would be so that the value of each option has a guarantee to match one of the date_term partials produced in the `render_edit_field_partial....` below. The JS that would drive this feature would detect the `change` event on this select, grab the value of the selected option and then find the related date_term partial below and make it visible.

wickr:
This looks good, and thanks for the screenshots. For the Search Results view, I don't think all of the Dates need to be included. I realize your test instance has all of them and most items won't, but there will likely be other fields to include as well. 

I would say to start with 1 Date in the search results view, probably Date Created.

wickr:
On the Item show page, under the Descriptions header, the field names only have the first word capitalized: "Date accepted" where the other instances have "Date Accepted". If that's a tweak that affects all of Hyrax then maybe that should get pushed up so it's consistent.

straleyb:
@luisgreg99 I had two questions, but after those are answered, I say we merge this puppy.

wickr:
Ranges should be stored _and_ viewed in the form as ISO 8601 format. Generally it's using a forward slash: "YYYY-MM-DD/YYYY-MM-DD"  or  "YYYY/YYYY" .  https://en.wikipedia.org/wiki/ISO_8601#Time_intervals
(you may be storing them this way, but the form is showing "01 - 04/06/2017"  and that's not a useful range)

How they are displayed on the show pages can vary, but there should be a method to handle this. There was a reference to a `human_readable_date` method in Hyrax somewhere, can't remember right now if that was a Rails or Hyrax method.

Range isn't really something to turn on/off per-date field, as all dates can be ranges, and you could type one in anyway, so I don't think that should be on the model. I think you can further simplify by just having one partial that can handle both single and range fields. The option can be there to add the end date, and if it's selected, use it, if not, ignore it.

If we do want certain dates to be *only* a single day or year value, we should look at validating the date value, not whether part of the UI is active or not.

luisgreg99:
@wickr I updated my branch to address your comment, here the changes I made on https://github.com/osulp/Scholars-Archive/pull/523/commits/ca0d5641f6233105c0a09dbd6587165d5f5d713a 
- updated date pikers and added the edtf gem to support date ranges in the ISO 8601 format: "YYYY-MM-DD/YYYY-MM-DD", which is what is also stored in the field
edit/new forms:
image
show page:
image

- added a method `human_readable_date_edtf` in the hyrax helper to parse date ranges in the ISO 8601 format and render a readable version, here's a screenshot:
image
- removed both the single and date range partials and made a `form_date_range_widget` partial that is enabled when the "Date range" checkbox is on
- removed the date_range_terms logic from the model and views so that any date term could be a range if the "Date range" checkbox is on

PR #554, opened 2017-05-02
Integrates CAS authentication for user creation.

### LINKED ISSUE #515: Setup CAS
Set.
It.
Up.

straleyb:
Fixes #515 

This adds CAS user authentication into Scholars Archive, with this people will be able to login via CAS with their onid username and it will create a user account for them.


revgum:
Instead of the previous application settings route.. can we use the local env.. Check out local_env and application.rb to see how the variables are injected into the app..? I feel like this would help keep our application config consistent.

PR #556, opened 2017-05-03
Construct page title for work and file stats page

### LINKED ISSUE #452: Provide page title value for Work and File Analytics pages, to match text in h2
Currently the page title for http://ir-staging.library.oregonstate.edu/files/8623hx72q/stats  is "File Stat // Sufia". "File Stat" should probably match "File Analytics" that is in the h2.

Because we don't provide a page_title value, I think Curation Concerns is just using a default page value from the controller name: https://github.com/projecthydra/curation_concerns/blob/71f0b9d8b210a7833eefdf333bfeb8859e255c56/app/helpers/curation_concerns/title_helper.rb#L18

zhang4952:
fix #452


app/views/hyrax/stats/file.html.erb
```diff
@@ -0,0 +1,24 @@
+<% provide :page_title, construct_page_title("File Analytics #{@stats.model.title.first}") %>
```
wickr: There should be punctuation of some sort, probably a ':' after File Analytics.

app/views/hyrax/stats/work.html.erb
```diff
@@ -0,0 +1,24 @@
+<% provide :page_title, construct_page_title("Work Analytics #{@stats.model.title.first}") %>
```
wickr: There should be punctuation of some sort, probably a ':' after Work Analytics.

PR #567, opened 2017-05-16
Feature/geo fields

### LINKED ISSUE #300: [Geo] Nested attributes for geographic bounding box
relates to #301 and #144

### LINKED ISSUE #455: Geographic Points should display Latitude and Longitude in addition to Label
For a Geographic Point field, only the Label is faceted and displayed on the show page, not the Latitude or Longitude values.

### LINKED ISSUE #457: Bounding Box should display points in addition to Label
For a Bounding Box field, only the Label is faceted and displayed on the show page, not the point values.

### LINKED ISSUE #460: Ensure that Geographic Point and Bounding Box do not appear in facets
Since these fields can only be faceted on the Labels that someone enters in, I don't know how useful it will be to facet these. If there's a strong reason to keep them, I would vote for having them very low in the list.

### LINKED ISSUE #514: UI Mockup for Date and Geo Picker

luisgreg99:
fixes #514 
fixes #460 
fixes #457 
fixes #455 
fixes #300 

**New Edit Form**

geo-fields

**Screenshots:**

- New/Edit form

image

- Show page:
image

- Search results view

image

- [x] Added facet fields and solr configuration for geo fields (points, bounding boxes)
- [x] Updated the presenter to render geo fields (points, bounding boxes) in the edit/new form
- [x] Added nested models for geo points bounding boxes attributes
- [x] Added proper js to handle dynamic add/remove actions for each geo field (points, bounding boxes, and geonames)


vantuyls:
looks good!

app/controllers/catalog_controller.rb
```diff
@@ -43,6 +43,8 @@ def self.modified_field
     #   The ordering of the field names is the order of the display
     config.add_facet_field solr_name("human_readable_type", :facetable), label: "Type", limit: 5
     config.add_facet_field solr_name("resource_type", :facetable), label: "Resource Type", limit: 5
+    config.add_facet_field solr_name("nested_geo_points_label", :symbol), label: "Geographic Point", limit: 5
```
wickr: Both of these should not be faceted, they'll be user-inputted labels and the chances of having more than one occurrence of the same value are slim.

app/controllers/catalog_controller.rb
```diff
@@ -80,6 +82,8 @@ def self.modified_field
     config.add_index_field solr_name("date_uploaded", :stored_sortable, type: :date), label: "Date Uploaded", itemprop: 'datePublished', helper_method: :human_readable_date
     config.add_index_field solr_name("rights", :stored_searchable), label: "Rights", helper_method: :license_links
     config.add_index_field solr_name("resource_type", :stored_searchable), label: "Resource Type", link_to_search: solr_name("resource_type", :facetable)
+    config.add_index_field solr_name("nested_geo_points_label", :stored_searchable), label: "Geographic Point", itemprop: 'point'
```
wickr: We don't need either of these to show up in the Search Results view either.

app/models/concerns/scholars_archive/default_metadata.rb
```diff
@@ -184,6 +184,11 @@ module DefaultMetadata
         index.as :stored_searchable
       end
 
+      property :nested_geo_points, :predicate => ::RDF::URI("http://www.w3.org/2003/12/exif/ns#geo"), :class_name => NestedGeoPoint
```
wickr: :nested_geo_points and :nested_geo_bbox can be combined into a single :nested_geo property. The predicate for that is https://purl.org/geojson/vocab#Feature

app/models/nested_geo_bbox.rb
```diff
@@ -0,0 +1,31 @@
+class NestedGeoBbox < ActiveTriples::Resource
+  property :label, predicate: ::RDF::Vocab::DC.coverage
```
wickr: All of the predicates have changed, so the new combined :nested_geo property should have these properties:

- label	dct:title
- point	https://purl.org/geojson/vocab#coordinates
- bbox	https://purl.org/geojson/vocab#bbox

config/locales/hyrax.en.yml
```diff
@@ -18,6 +18,10 @@ en:
           date_accepted: "Date Accepted"
           date_collected: "Date Collected"
           date_valid: "Date Valid"
+      form:
+        nested_geo_points: "Geographic Points"
+        nested_geo_bbox: "Bounding Box"
+        nested_geo_location: "Geographic Location"
```
wickr: This line can go away since :nested_geo_location is gone.

wickr:
Another thing we need to address is on the Show page, or any time these fields are displayed, we need to show both the label and the coordinates. It looks like right now only the label is stored in Solr. I'm not sure if we add the coordinates to Solr in a separate field, if there were then more than 1 pair if you could tell them apart. Another option is to append the coordinates to the label so it has everything to show:  `United States  [##.##, ###.##, ##.##, ###.##]`  This may suffice for showing the values, we would need to do something else to take action on the coordinates directly for future mapping purposes.

This is similar to the issue for linked data storing in Solr and showing label and URI and needing them to stay connected.

@revgum @straleyb @vantuyls @zhang4952

PR #573, opened 2017-05-17
Adds replaces to edit form and show page. Adds rake task for mapping

### LINKED ISSUE #562: work ID => handle mapping rake task
Generate a rake task that can query for a mapping from work ID to its handle. The resulting map will be used to configure nginx to rewrite urls that are pointing at old DSpace handle links.

### LINKED ISSUE #571: expose replaces field on work form and presenters
The field is added, but inaccessible or viewable

straleyb:
Fixes #571 Fixes #562

### DIFF
--- a/app/presenters/etd_presenter.rb
+++ b/app/presenters/etd_presenter.rb
@@ -1,4 +1,3 @@
-# app/presenters/default_work_presenter.rb
-class DefaultWorkPresenter < Hyrax::WorkShowPresenter
+class EtdPresenter < Hyrax::WorkShowPresenter
   delegate :date_accepted, :date_available, :date_collected, :date_copyright, :date_issued, :replaces, :date_valid, to: :solr_document
 end
--- a/app/views/hyrax/base/_attribute_rows.html.erb
+++ b/app/views/hyrax/base/_attribute_rows.html.erb
@@ -12,7 +12,7 @@
 <%= presenter.attribute_to_html(:language, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:identifier, render_as: :linked, search_field: 'identifier_tesim') %>
 <%= presenter.attribute_to_html(:keyword, render_as: :faceted) %>
-
+<%= presenter.attribute_to_html(:replaces, render_as: :external_link) %>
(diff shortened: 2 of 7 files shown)


lib/tasks/handle_dump.rake
```diff
@@ -0,0 +1,12 @@
+namespace :scholars_archive do
+  desc "Dump all handles associated with all objects in db"
+  task :handle_dump do
+    puts "Beginning building"
+    @hash = {}
+    response = ActiveFedora::SolrService.get("replaces_tesim:*", :rows => 10000)["response"]["docs"]
```
wickr: Rows needs to be higher than 10000, I'd probably do 100,000

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -12,7 +12,7 @@
 <%= presenter.attribute_to_html(:language, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:identifier, render_as: :linked, search_field: 'identifier_tesim') %>
 <%= presenter.attribute_to_html(:keyword, render_as: :faceted) %>
-
+<%= presenter.attribute_to_html(:replaces, render_as: :faceted) %>
```
wickr: Replaces is not faceted, external link is probably fine

straleyb: I didnt think the handles were going to be on the show page or the form. Is that right? If so, Im not so sure this change is totally necessary.

app/presenters/etd_presenter.rb
```diff
@@ -0,0 +1,3 @@
+class ETDPresenter < Hyrax::WorkShowPresenter
```
revgum: The convention seems to be EtdPresenter, you should rename this and the references to it.

straleyb:
Is it ok if I alpha sort the metadata in another PR? I have that going along with adding in the rest of the metadata from the map.

revgum:
@straleyb ah hah, my mistake.. carry on with that other PR for alpha sorting

straleyb:
Ok updated @revgum 👍

PR #610, opened 2017-06-01
Adds two new models with forms and show pages working properly.

straleyb:
This needs to be checked out. There may be some useless files here and maybe some discussion could be had about making the views/hyrax/base/_attribute_rows partial look a bit nicer. Also the solr_document is getting a bit longer now. Im working on writing specs for these and the other models that exist in hyrax.


app/models/concerns/scholars_archive/default_metadata.rb
```diff
@@ -184,6 +184,11 @@ module DefaultMetadata
         index.as :stored_searchable
       end
 
+      id_blank = proc { |attributes| attributes[:id].blank? }
+
+      class_attribute :controlled_properties
+      self.controlled_properties = [:based_near]
+      accepts_nested_attributes_for :based_near, reject_if: id_blank, allow_destroy: true
```
revgum: I'd probably just put the proc directly in the attribute until/unless it is needed/used elsewhere

app/models/concerns/scholars_archive/oer_metadata.rb
```diff
@@ -0,0 +1,27 @@
+module ScholarsArchive
+  module OerMetadata
+    extend ActiveSupport::Concern
+
+    included do
+      property :time_required, predicate: ::RDF::URI.new("http://purl.org/dcx/lrmi-terms/timeRequired") do |index|
```
revgum: can we alpha sort the properties?

app/models/concerns/scholars_archive/article_metadata.rb
```diff
@@ -0,0 +1,44 @@
+module ScholarsArchive
+  module ArticleMetadata
+    extend ActiveSupport::Concern
+
+    included do
+      #reusable metadata fields for DSpace migration
```
revgum: can we alpha sort the properties?

app/models/solr_document.rb
```diff
@@ -149,4 +149,64 @@ def degree_name
   def graduation_year
     self[Solrizer.solr_name('graduation_year')]
   end
+
```
revgum: something worth considering would be to take an array of the field names and dynamically define these standard methods for them all at once;

```
%w(time_required typical_age_range learning_resource_type).each do |property_name|
  define_method property_name do
    self[Solrizer.solr_name(property_name)]
  end
end
```

straleyb: Yeah, I did something like that in the tests that I just pushed up. Id definitely be willing to simplify this a bit. 👍

app/models/article.rb
```diff
@@ -0,0 +1,15 @@
+# Generated via
+#  `rails generate hyrax:work Article`
+class Article < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::ScholarsArchive::ArticleMetadata
+  include ::ScholarsArchive::DefaultMetadata
+  include ScholarsArchive::TriplePoweredProperties::WorkBehavior
+  include SchoalrsArchive::ToSolrBehavior
```
revgum: `ScholarsArchive` is misspelled here.

app/models/default_work.rb
```diff
@@ -4,27 +4,10 @@ class DefaultWork < ActiveFedora::Base
   include ::Hyrax::WorkBehavior
   include ::ScholarsArchive::DefaultMetadata
   include ::ScholarsArchive::TriplePoweredProperties::WorkBehavior
+  include SchoalrsArchive::ToSolrBehavior
```
revgum: `ScholarsArchive` is misspelled here.

app/models/etd.rb
```diff
@@ -6,27 +6,11 @@ class Etd < ActiveFedora::Base
   include ::ScholarsArchive::EtdMetadata
   include ::ScholarsArchive::DefaultMetadata
   include ScholarsArchive::TriplePoweredProperties::WorkBehavior
+  include SchoalrsArchive::ToSolrBehavior
```
revgum: `ScholarsArchive` is misspelled here.

app/models/oer.rb
```diff
@@ -0,0 +1,15 @@
+# Generated via
+#  `rails generate hyrax:work Oer`
+class Oer < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::ScholarsArchive::OerMetadata
+  include ::ScholarsArchive::DefaultMetadata
+  include ScholarsArchive::TriplePoweredProperties::WorkBehavior
+  include SchoalrsArchive::ToSolrBehavior
```
revgum: `ScholarsArchive` is misspelled here.

app/models/to_solr_behavior.rb
```diff
@@ -0,0 +1,23 @@
+module ToSolrBehavior
```
revgum: I think this should be deleted, it seems you have the appropriate one in `concerns/scholars_archive`

app/models/concerns/scholars_archive/to_solr_behavior.rb
```diff
@@ -0,0 +1,25 @@
+module ScholarsArchive
+  module ToSolrBehavior
```
revgum: I feel like this would be better named relative to what it is doing, `HasSolrNestedGeoLabels`.. what do you think?

app/forms/hyrax/date_terms_behavior.rb
```diff
@@ -0,0 +1,11 @@
+module Hyrax
```
revgum: I think this is something we should namespace to `ScholarsArchive` instead of `Hyrax`

app/models/concerns/scholars_archive/to_solr_behavior.rb
straleyb: Yeah I like this. to_solr is a little indirect and I wouldnt quite know what it does. But that name describes what the to_solr method override is doing. 👍

PR #612, opened 2017-06-05
Feature/add higher level models

straleyb:
This adds in the models from the spreadsheet. Still need to test the presenters and the forms.

### DIFF
--- a/app/models/conference_proceedings_or_journal.rb
+++ b/app/models/conference_proceedings_or_journal.rb
@@ -1,0 +1,18 @@
+# Generated via
+#  `rails generate hyrax:work ConferenceProceedingsOrJournal`
+class ConferenceProceedingsOrJournal < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::ScholarsArchive::ArticleMetadata
+  include ::ScholarsArchive::DefaultMetadata
+  include ScholarsArchive::TriplePoweredProperties::WorkBehavior
+  include ScholarsArchive::HasSolrNestedGeoLabels
+
+  self.indexer = ArticleIndexer
+
+
+  # Change this to restrict which works can be added as a child.
+  # self.valid_child_concerns = []
+  validates :title, presence: { message: 'Your work must have a title.' }
+
+  self.human_readable_type = 'Conference Proceedings Or Journal'
+end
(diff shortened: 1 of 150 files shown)


app/models/conference_proceedings_or_journal.rb
```diff
@@ -0,0 +1,15 @@
+# Generated via
+#  `rails generate hyrax:work ConferenceProceedingsOrJournal`
+class ConferenceProceedingsOrJournal < ActiveFedora::Base
+  include ::Hyrax::WorkBehavior
+  include ::ScholarsArchive::ArticleMetadata
+  include ::ScholarsArchive::DefaultMetadata
+  include ScholarsArchive::TriplePoweredProperties::WorkBehavior
+
+  self.indexer = ConferenceProceedingsOrJournalIndexer
```
revgum: Might be worth considering we eliminate an indexer for each of these models, and make indexers that are composed in a way to handle the mix of metadata that our models use. For instance, this model just needs an indexer that handles indexing both `ArticleMetadata` and `DefaultMetadata`.. whereas other models only need `DefaultMetadata`, and a couple have both `EtdMetadata` and `DefaultMetadata`. I think we could get away with 4 Indexers,

`DefaultIndexer`, `EtdIndexer`, `OerIndexer`, and `ArticleIndexer`.

spec/actors/hyrax/actors/administrative_report_or_publication_actor_spec.rb
```diff
@@ -0,0 +1,9 @@
+# Generated via
+#  `rails generate hyrax:work AdministrativeReportOrPublication`
+require 'rails_helper'
+
+RSpec.describe Hyrax::Actors::AdministrativeReportOrPublicationActor do
```
revgum: I think we should remove these since the actors are all just inheriting from base.. we don't need bogus specs hanging out doing nothing.

spec/models/administrative_report_or_publication_spec.rb
```diff
@@ -0,0 +1,9 @@
+# Generated via
```
revgum: A good start for covering the models would be to check their indexers and human readable type.

app/models/conference_proceedings_or_journal.rb
straleyb: https://github.com/samvera-labs/hyrax/blob/68527862df498aa9f407b909d4f41d781ba146dc/app/indexers/hyrax/basic_metadata_indexer.rb#L5-L7

PR #614, opened 2017-06-05
added work url in the show page

### LINKED ISSUE #520: show the 'canonical'/citable URI alongside the descriptive metadata
URLs get all junked up with stuff, we want the clean URI in the descriptive metadata on the show page for each item.

luisgreg99:
Added citeable url for the current work in the show page.

fixes #520 

Screenshot:
image


app/views/hyrax/base/_work_description.erb
```diff
@@ -0,0 +1,6 @@
+<label><%= t('hyrax.work_url_label') %>:</label>
+<span><%= Rails.application.routes.url_helpers.url_for(:only_path => false, :action => 'show', :controller => 'hyrax/'+presenter.model_name.plural,  :host=>request.host, id: presenter.id) %></span>
```
wickr: Can you clean up the spacing here around :host?

wickr:
Looks good other than my 1 comment. I know it's small but I think it should have a view spec.

PR #628, opened 2017-06-09
Feature/default metadata updates

luisgreg99:


app/controllers/catalog_controller.rb
```diff
@@ -105,7 +112,9 @@ def self.modified_field
     config.add_show_field solr_name("date_modified", :stored_searchable), label: "Date Modified"
     config.add_show_field solr_name("date_uploaded", :stored_searchable), label: "Date Uploaded"
     config.add_show_field solr_name("date_valid", :stored_searchable), label: "Date Valid"
+    config.add_show_field solr_name("hydrologic_unit_code", :stored_searchable), label: "Hydrologic Unit Code"
```
straleyb: Arent these showing up on the single result show field already? Via the presenter override.

app/models/concerns/scholars_archive/default_metadata.rb
```diff
@@ -8,6 +8,14 @@ module DefaultMetadata
         index.as :stored_searchable
       end
 
+      property :academic_affiliation, predicate: ::RDF::URI("http://vivoweb.org/ontology/core#AcademicDepartment") do |index|
+        index.as :stored_searchable, :facetable
+      end
+
+      property :other_affiliation, predicate: ::RDF::URI("http://vivoweb.org/ontology/core#Department") do |index|
```
revgum: Can we keep these properties alpha sorted, moving `other_affiliation` down?

PR #646, opened 2017-06-14
Fixes location saving

straleyb:
Im still trying to figure out why two fields are rendering. It seems as though it has something to do with https://github.com/samvera/hyrax/blob/master/app/views/records/edit_fields/_based_near.html.erb#L2

When you remove the as: :controlled_vocabulary completely it only renders one field. Otherwise it renders two.


app/forms/scholars_archive/nested_geo_behavior.rb
```diff
@@ -0,0 +1,20 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      delegate :nested_geo_attributes=, :to => :model
+      def initialize_fields
+        model.nested_geo.build
+        super
+      end
+      def self.build_permitted_params
+        super + self.date_terms + [
```
revgum: `self.date_terms` seems like it shouldn't be in this behavior

PR #649, opened 2017-06-15
Edit sa_gtd_workflow.json

vantuyls:
edited workflow step names to make them more readable in the UI


vantuyls:
@revgum PR here with an update to the GTD Workflow. 

Want to see if you can update the workflow without blowing up all the stuff?

config/workflows/sa_gtd_workflow.json
```diff
@@ -73,10 +73,10 @@
                         }
                     ]
                 }, {
-                    "name": "comment_only",
+                    "name": "Comment_Only",
```
revgum: I wonder if these names can do without the underscores and have clean looking meaningful names throughout. Should we try?

PR #665, opened 2017-06-26
Feature/decade date facet

### LINKED ISSUE #584: Add Date facet with Range slider
Look at adding: https://github.com/projectblacklight/blacklight_range_limit

This needs a Solr integer field with a year, probably 'date_year_facet'. Priority for populating this date field is: Date Created, Date Copyright, Date Issued

### LINKED ISSUE #585: Add Decade date facet
In conjunction with #584, this will be a facet of decades: "1910-1919, 1920-1929, etc."

Probably called 'date_decade_facet'.

For a single date to use for an item, priority is: Date Created, Date Copyright, Date Issued

Can look at how this was done in Oregon Digital: https://github.com/OregonDigital/oregondigital/blob/master/lib/oregon_digital/date_sorting.rb

luisgreg99:
Fixes #584 
Fixes #585 

Screenshot:

image

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -1,4 +1,5 @@
 class CatalogController < ApplicationController
+  include BlacklightRangeLimit::ControllerOverride
   include Hydra::Catalog
   include Hydra::Controller::ControllerBehavior
 
@@ -21,7 +22,7 @@ class CatalogController < ApplicationController
 
     config.show.tile_source_field = :content_metadata_image_iiif_info_ssm
     config.show.partials.insert(1, :openseadragon)
-    config.search_builder_class = Hyrax::CatalogSearchBuilder
+    config.search_builder_class = ScholarsArchive::CatalogSearchBuilder
 
     # Show gallery view
     config.view.gallery.partials = [:index_header, :index]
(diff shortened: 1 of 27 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -68,6 +69,8 @@ def self.modified_field
     config.add_facet_field solr_name('degree_level', :facetable), limit: 5, label: 'Degree Level'
     config.add_facet_field solr_name('degree_name', :facetable), limit: 5, label: 'Degree Name'
     config.add_facet_field solr_name('graduation_year', :facetable), limit: 5, label: 'Commencement Year'
+    config.add_facet_field('date_decades_ssim', :label => 'Decade', :limit => 10, :sort => 'desc')
```
wickr: Think you can drop the parentheses here.

app/models/concerns/scholars_archive/date_operations.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      def last_year
+        year + 10 - (year+10)%10 - 1
+      end
+    end
+
+    class DateDecadeConverter
+      attr_accessor :date
+      def initialize(date)
```
revgum: It would be nice to have a comment to show expected examples of the value of `date` to help understand how the other related methods are processing.

app/models/concerns/scholars_archive/date_operations.rb
```diff
@@ -0,0 +1,134 @@
...
+    end
+
+    def clean_years
+      date = Date.edtf(date_value)
+      if date.instance_of? EDTF::Interval
+        date.map(&:year).uniq
+      elsif date.instance_of? Date
+        Array.wrap(date.year)
+      else
+        Array.wrap(clean_datetime.year)
```
revgum: Because in both places you're using `clean_datetime.year`, it seems to be more clear if the method would return the year. For instance, 

`def parsed_year`

app/models/concerns/scholars_archive/date_operations.rb
```diff
@@ -0,0 +1,134 @@
...
+      def valid_date_range?
+        dates.first.to_s.length == 4 && dates.last.to_s.length == 4 && dates.length == 2
+      end
+
+      def dates
+        @dates ||= date.to_s.split("-").map(&:to_i)
+      end
+
+      def decades
+        if calculated_decades <= 3
```
revgum: I think you could combine `decades` and `calculated_decades` to just be, since `calculated_decades` the method is isolated to being used only within `decades`

```
def decades
  calculated_decades = (dates.last - dates.first)/10 + 1
  calculated_decades <= 3 ? calculated_decades : 0
end
```

PR #693, opened 2017-06-30
Feature/search and external link renderer

### LINKED ISSUE #640: URIs on Showpage should behave like...
...
- [ ] The label link should activate a faceted search

- [ ] Add an external link icon that takes you to the URI landing page

luisgreg99:
image

fixes #640

### DIFF
--- a/app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb
@@ -1,0 +1,27 @@
+module Hyrax
+  module Renderers
+    class SearchAndExternalLinkAttributeRenderer < AttributeRenderer
+      include ApplicationHelper
+      private
+
+      def li_value(value)
+        links_to_search_field_and_external_uri(search_field, value)
+      end
+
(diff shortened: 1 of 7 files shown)


app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb
```diff
@@ -0,0 +1,40 @@
...
+        search_path = Rails.application.class.routes.url_helpers.search_catalog_path(
+            search_field: search_field, q: ERB::Util.h(query)
+        )
+        links << link_to(query, search_path)
+        links << link_to('<span class="glyphicon glyphicon-new-window"></span>'.html_safe, uris.first, 'aria-label' => "Open link in new window", class: 'btn btn-defaul') if uris.first.present?
+        links.join('')
+      end
+
+      def label_uris
+        ScholarsArchive::TriplePoweredService.new.fetch(uri_presenter)
```
revgum: The label/uri are index in SOLR, so it seems like we would want to use/parse those instead of making extra calls to the TriplePoweredService to fetch the label again.

PR #702, opened 2017-07-05
Refactors duplicate code into behaviors

### LINKED ISSUE #689: Eliminate unneeded models, forms, controllers, etc
`DefaultWork`, `Etd`, and `Oer` models and related forms could be reworked as behaviors rather than extra/confusing code

straleyb:
Fixes #689


straleyb:
@revgum So one thing im having trouble with here is including both the `DefaultWorkFormBehavior` and another `*WorkFormBehavior` causes the default values to be overridden. Calling super wont go through the chain of `self.terms` and we are dropping fields.

straleyb:
@revgum Im pretty sure that was done here. Am I missing views? https://github.com/osulp/Scholars-Archive/pull/702/files#diff-dc6efe9cdc5491e205b74b3bf9ae2066L1

PR #722, opened 2017-07-10
added mailboxer_email method for messaging

### LINKED ISSUE #707: Configure/enable notification emails
allow SA@OSU to send emails as notifications, in addition to the in-system notification panel.

luisgreg99:
fixes #707

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+require 'rails_helper'
+RSpec.describe User, type: :model do
+  let(:user) do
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  it "has an email" do
+    expect(user.email).to be_kind_of String
+    expect(user.email).to eq ("test@example.com")
+  end
+
+  it "has mailbox-related methods defined" do
+    expect(user).to respond_to(:mailboxer_email)
+    expect(user).to respond_to(:name)
```
revgum: I think one more, which test the return value expected `expect(user.mailboxer_email).to eq user.email`

PR #724, opened 2017-07-11
added other input for degree_field and degree_level select

### LINKED ISSUE #719: Fields that should have the Other option
- [ ] Degree Name
- [ ] Degree Field

### LINKED ISSUE #727: Local database table for tracking entries to "Other" fields
See #719 

- [ ] Create an Actor which is responsible for persisting the work_id and the values for the "Other" fields in a local table during "create". During "update" add/remove "Other" field values from the row for this work. This enables the ability to persist these values outside of Fedora and SOLR when a user intends on asking for a new degree name, or field to be added to the system.

- [ ] Add ability to fetch the "Other" fields values from the database for a work

luisgreg99:
image

- Added add_other js and updated the form partials to allow the user to enter custom values when selecting "Other" for "degree_field" and "degree_level": fixes #719

- Implemented the `ScholarsArchive::Actors::AddOtherFieldOptionActor` that is used to persist "other" values provided by the user in the form into the `OtherOption` table: closes #727

### DIFF
--- a/app/controllers/concerns/scholars_archive/works_controller_behavior.rb
+++ b/app/controllers/concerns/scholars_archive/works_controller_behavior.rb
@@ -9,10 +9,35 @@ module ScholarsArchive
 
     def edit
       parse_geo
+      get_other_option_values
+      super
+    end
+
+    def update
+      set_other_option_values
+      super
+    end
+
+    def create
+      set_other_option_values
       super
     end
 
     private
+
+    def set_other_option_values
+      # if the user selected the "Other" option in "degree_field" or "degree_level", and then provided a custom
+      # value in the input shown when selecting this option, these custom values would be assigned to the
+      # "degree_field_other" and "degree_level_other" attribute accessors so that they can be accessed by
+      # AddOtherFieldOptionActor. This actor will persist them in the database for reviewing later by an admin user
+      if params[hash_key_for_curation_concern]['degree_field'] == 'Other' && params[hash_key_for_curation_concern]['degree_field_other'].present?
+        curation_concern.degree_field_other = params[hash_key_for_curation_concern]['degree_field_other']
+      end
+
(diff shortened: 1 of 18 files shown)


app/controllers/concerns/scholars_archive/works_controller_behavior.rb
```diff
@@ -12,8 +12,32 @@ def edit
       super
     end
 
+    def update
+      set_other_option_values
```
revgum: I feel like both #update and #create would be better suited as `before_action` callbacks and to set_other_option_values?

revgum:
Did you look into if it’s possible to configure hyrax (maybe in the initializer.. or somewhere low level) to use a namespaced `ScholarsArchive::DefaultMiddlewareStack` that inherits from `Hyrax::DefaultMiddlewareStack` and injects our actor into and along with the others that are provided by the Hyrax version.. that way we don’t have a manual override to keep in sync with master?

app/controllers/concerns/scholars_archive/works_controller_behavior.rb
luisgreg99: @revgum yeah I tried calling it with `:before_action`, but for some reason `curation_concern` is nil when the method is ran

config/initializers/engine.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
```
revgum: I think we want to find a way not override the Hyrax::Engine, but to just override the `config.to_prepare` block.

app/controllers/concerns/scholars_archive/works_controller_behavior.rb
```diff
@@ -12,8 +12,32 @@ def edit
       super
```
revgum: Would you want to fetch the appropriate field & value from the database during `#edit` as well?

config/initializers/engine.rb
luisgreg99: It looks like setting it in application.rb works as well, I think that would be another way

PR #735, opened 2017-07-13
Feature/auto admin set

### LINKED ISSUE #662: Use Work Type to define admin set automatically
Work type should define the admin set in the relationships tab of the input form for a new work.

straleyb:
Fixes #662


app/services/scholars_archive/admin_set_select_service.rb
```diff
@@ -0,0 +1,9 @@
+module ScholarsArchive
+  class AdminSetSelectService
+    def self.select(model, select_options)
+      admin_set_name = YAML.load(File.read("config/admin_set_map.yml"))[model]
+      mapped_admin_set = select_options.select{ |option| option.first == admin_set_name }.flatten
```
revgum: #casecmp seems safe because it always ends up that configuration files and string comparison ends up being code smell. 


```
mapped_admin_set = select_options.find { |o| option.first.casecmp(admin_set_name).zero? }
[ mapped_admin_set || ENV["SCHOLARSARCHIVE_DEFAULT_ADMIN_SET"] ]
```

config/admin_set_map.yml
```diff
@@ -0,0 +1,14 @@
+#This file is in the order of `ModelName: "Admin Set Name"`
+#If changes are being made to this file, please follow this order
+
+Article: "Article"
+Graduate Thesis Or Dissertation: "Graduate Thesis Or Dissertation"
```
revgum: The keys are all humanized.. you want to remove the spaces to map them as models.

app/services/scholars_archive/admin_set_select_service.rb
```diff
@@ -0,0 +1,9 @@
+module ScholarsArchive
+  class AdminSetSelectService
+    def self.select(model, select_options)
```
revgum: I think this method would benefit from a brief comment about what it's doing.. I re-read it this morning and took a bit to jog my memory of how this relates to the select html element, and what this intends on returning.

PR #738, opened 2017-07-14
Adds honors college model

### LINKED ISSUE #672: Build Admin Set and Work Type for Honors College Theses

straleyb:
Fixes #672 Just need to add admin set on prod server.


revgum:
see #739

Issue #720, opened 2017-07-10
Enable search for academic units in form


vantuyls:
need typeahead for academic units in the form

vantuyls:
want to make this work from the ONS jsonld and use rails cache store.

straleyb:
Per our discussion. Uses `ServiceBasedAuthority` which hits cached JSON-LD and parses it similar to `FileBasedAuthority` to make typeahead do its thing.

vantuyls:
@straleyb is this completed? i'm not seeing this functionality in SA@OSU right now, but maybe i'm missing something.

vantuyls:
@simholt any sense for how important it seems to have typeahead on academic units?

simholt:
@vantuyls 

Academic units = all the degree fields and things? 

Not super-important. It would be nice but the drop-downs function fine.

lamtu1:
@carakey @simholt Would it be okay if I can do a quick tackle to get this ticket out of the way. Since I deal with the `ROR` functionality, I can get this one in for academic easily.

simholt:
Fine with me!

simholt:
This isn't working -- will give detailed feedback!

lamtu1:
NOTE: This PR got reverted from last time so I need to check down and see how to get it back in.

lamtu1:
Ready to QA: Need to look into a way to reindex work with `academic affiliation`

simholt:
- Remove URLs from the drop-down.
- Need to trim the width so it doesn't take over the screen on the right side. (Likely not a problem if URLs moved but still should be done.)
- Can we make it so deprecated terms do not show up? Example: https://opaquenamespace.org/ns/osuAcademicUnits/hbrEnAJQ

<img width="890" alt="Image" src="https://github.com/user-attachments/assets/5de2ff79-5b46-42c2-9567-9abfec434bd2" />

lamtu1:
`FUTURE NOTE:` For updating the `prod` once everything is move to `prod` later after staging

```
r = ActiveFedora::SolrService.query("academic_affiliation_label_tesim:*", fl: 'id', rows: 10000)
r.map { |x| x['id'] }.each_with_index do |pid, index|
  w = ActiveFedora::Base.find(pid)
  vals = w.academic_affiliation
  w.academic_affiliation = []
  vals.each do |i|
    w.academic_affiliation << ScholarsArchive::ControlledVocabularies::AcademicAffiliation.new(i)
  end
[...]

PR #741, opened 2017-07-18
Adds caching support for configurable and default vocabulary

### LINKED ISSUE #725: [ONS Authority] Service
Create a service which fetches the JSONLD of a vocabulary, and uses Rails caching mechanisms to keep it fresh. The service maps the JSONLD to the data shape expected by QA (id, label, active).

### LINKED ISSUE #726: [ONS Authority] ServiceBasedAuthority
It seems like using QA FileBasedAuthority as a foundation for implementing a ServiceBasedAuthority.. https://github.com/samvera/questioning_authority/blob/master/lib/qa/authorities/local/file_based_authority.rb

Instead of parsing a YAML file, it could use a service which is tied to ONS and uses Rails cache for fetching the jsonld and mapping it to the proper form for QA terms (id, label, active). 

See #725

straleyb:
Fixes #725 
Fixes #726 
References #720


app/authorities/parsers/academic_units_parser.rb
```diff
@@ -0,0 +1,13 @@
+module Parsers
+  class AcademicUnitsParser
+    def self.parse(jsonld)
+      extract_labels(jsonld)
+    end
+
+    def self.extract_labels(jsonld)
+      labels = JSON.parse(jsonld).dig('@graph').map { |graph| graph.dig("rdfs:label").dig("@value") unless graph.dig("rdfs:label").nil? }
```
revgum: `#dig` takes an arbitrarily deep nested hash and returns nil if the any of the keys don't exist. Also, for readability sake I feel like this is better:

```
graph = JSON.parse(jsonld).dig("@graph")
raise "Did not find @graph in jsonld" unless graph
terms = graph.map { |g| { id: g.dig("@id"), label: g.dig("rdfs:label", "@value"), active: true } }
# return only terms that have id and label
terms.delete_if { |k,v| k.nil? || v.nil? } 
```

app/controllers/application_controller.rb
```diff
@@ -12,6 +12,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   before_action :check_d2h_http_header_auth
+
```
revgum: odd, not necessary

app/authorities/scholars_archive/cache_based_authority.rb
```diff
@@ -0,0 +1,20 @@
+module ScholarsArchive
+  class CacheBasedAuthority < Qa::Authorities::Local::FileBasedAuthority
+    private
+
+      def terms
+        authority_yaml = YAML.load(File.read(subauthority_filename))
+        uri = authority_yaml.with_indifferent_access.dig(@subauthority, :uri)
+        raise "#{@subauthority} uri configuration not found. " unless uri
+        result = ScholarsArchive::CachingService.fetch_or_store_in_cache(uri)
```
revgum: would be ideal to grab the cache timeout from the config here and pass into the CachingService

app/services/scholars_archive/caching_service.rb
```diff
@@ -0,0 +1,21 @@
+require 'net/http'
+
+module ScholarsArchive
+  class CachingService
+    def self.fetch_or_store_in_cache(vocabulary)
+      Rails.cache.fetch("/caching/#{vocabulary}", expires_in: 1.day, race_condition_ttl: 15.seconds) do
+        get_jsonld(vocabulary)
```
revgum: I like the 1.day expiration, seems like the expiration should be an argument to this method instead.. possibly the race condition too.

Since CachingService could be generic and fetch URIs, you could change `vocabulary` to `uri`, and `#get_jsonld` to `#get_response` .. that way the class is general and clear with its responsibilities.

revgum: ...also, the ".jsonld" should be coming in from the configuration.. this class would benefit from expecting full URIs and no hard coding.

config/environments/development.rb
```diff
@@ -17,16 +17,15 @@
 
   # Enable/disable caching. By default caching is disabled.
   if Rails.root.join('tmp/caching-dev.txt').exist?
+    puts "Config!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!1"
```
revgum: not needed

app/services/scholars_archive/caching_service.rb
```diff
@@ -0,0 +1,21 @@
+require 'net/http'
+
+module ScholarsArchive
+  class CachingService
+    def self.fetch_or_store_in_cache(vocabulary)
+      Rails.cache.fetch("/caching/#{vocabulary}", expires_in: 1.day, race_condition_ttl: 15.seconds) do
```
revgum: I believe `#fetch` is taking a string as the key.. if you verify that, then the key should just be the argument `uri` .. that way there isn't an unintended implied relationship between this key and the `config.cache_store`

app/authorities/parsers/academic_units_parser.rb
```diff
@@ -0,0 +1,15 @@
+module Parsers
+  class AcademicUnitsParser
+    def self.parse(jsonld)
+      extract_and_massage_data(jsonld)
+    end
+
+    private
+
+      def self.extract_and_massage_data(jsonld)
+        labels = JSON.parse(jsonld).dig('@graph').map { |graph| graph.dig("rdfs:label").dig("@value") unless graph.dig("rdfs:label").nil? }
```
revgum: `#dig` takes an arbitrarily deep nested hash and returns nil if the any of the keys don't exist. Also, for readability sake I feel like this is better:

```
graph = JSON.parse(jsonld).dig("@graph")
raise "Did not find @graph in jsonld" unless graph
terms = graph.map { |g| { id: g.dig("@id"), label: g.dig("rdfs:label", "@value"), active: true } }
# return only terms that have id and label
terms.delete_if { |k,v| k.nil? || v.nil? } 
```

app/services/scholars_archive/caching_service.rb
```diff
@@ -0,0 +1,22 @@
+require 'net/http'
+
+module ScholarsArchive
+  class CachingService
+    def self.fetch_or_store_in_cache(vocabulary, expires_in)
+      Rails.cache.fetch("/caching/#{vocabulary}", expires_in: expires_in, race_condition_ttl: 15.seconds) do
```
revgum: I believe #fetch is taking a string as the key.. if you verify that, then the key should just be the argument uri .. that way there isn't an unintended implied relationship between this key and the config.cache_store

config/authorities.yml
```diff
@@ -1 +1,5 @@
-:local_path: "config/authorities"
\ No newline at end of file
+:local_path: "config/authorities"
+academic_units:
+  uri: 'http://opaquenamespace.org/ns/osuAcademicUnits.jsonld'
+  cache_expires_in: 10.minutes
```
revgum: I kinda like the default of `1.day` or `12.hours` or so.. that way our app isn't query ONS a whole lot because in general that vocabulary isn't going to get updated.

straleyb:
screen shot 2017-07-19 at 3 11 17 pm

straleyb:
So here is what we currently have for the academic_affiliation dropdown using the caching service. As you can see there are some things that need to be addressed. It looks like there are some repeat values that are coming in from opaquenamespace. Are these unique? Do we need to keep these in and disambiguate them? Should we remove repeat values? (Depending on if their URIs are the same.) So far its looking good and can be extended it pretty easily to inspect the date values and find only ones used in the past 5 years as well.

PR #743, opened 2017-07-18
updated create method on AddOtherFieldOptionActor

### LINKED ISSUE #742: AddOtherFieldOptionActor not returning true on create
Found that an actor is expected to return true on the create and update methods to allow the stack to continue to the next actor and finalize the process events successfully in the actor chain/stack.

luisgreg99:
fixes #742


spec/actors/scholars_archive/actors/add_other_field_option_actor_spec.rb
```diff
@@ -25,40 +25,41 @@
   describe '#create' do
     context 'with other values selected for degree_field and degree_level' do
       let(:attributes) { { title: ["test"], creator: ["Blah"], rights_statement: ["blah.blah"], resource_type: ["blah"], degree_field: "Other", degree_level: "Other" } }
+      let(:test_degree_level_other) { "test1 degree level other" }
+      let(:test_degree_field_other) { "test1 degree level other" }
```
wickr: Looks like the text here should include 'field' to match the name

PR #754, opened 2017-07-20
Feature/dates for academic units

straleyb:


straleyb:
Need to fix tests and also curious about what we should do about the data inconsistencies/multiple dates per single label.

app/authorities/parsers/academic_units_parser.rb
```diff
@@ -9,9 +9,23 @@ def self.parse(jsonld)
...
         graph = JSON.parse(jsonld).dig("@graph")
         return [{ id: "invalid", term: "invalid", active: true }] if graph.nil?
-        terms = graph.map { |g| { id: g.dig("@id"), term: g.dig("rdfs:label", "@value"), active: true } }
+        terms = graph.map { |g| { id: g.dig("@id"), term: labels_with_dates(g), active: true } }
         # return only terms that have id and label
         terms.delete_if { |term| term[:id].nil? || term[:term].nil? }
       end
+
+      def self.labels_with_dates(g)
+        if g.dig("rdfs:label", "@value").nil?
```
revgum: This could be optimized for readability
```
label = g.dig("rdfs:label", "@value")
return label if label.nil?
date = g.dig("dc:date", "@value")
date = g.dig("dc:date") unless date
date = date.join(", ") if date.is_a?(Array)
"#{label} - #{date}"

PR #755, opened 2017-07-21
set proper recipients in workflow configs and email notification objects

luisgreg99:
- Updated `ChangesRequiredNotification` and `DepositReceivedNotification` to send to the user who initiated deposit only
- Updated `PendingReviewNotification` to send the emails to the approving role only (the user who made the deposit should get the `DepositReceivedNotification` email instead)
- Added diagrams for One Step and Graduate Thesis and Dissertation Workflows: https://github.com/osulp/Scholars-Archive/wiki/Workflows

### DIFF
--- a/app/services/scholars_archive/workflow/pending_review_notification.rb
+++ b/app/services/scholars_archive/workflow/pending_review_notification.rb
@@ -12,7 +12,7 @@ module ScholarsArchive
       end
 
       def users_to_notify
-        super << user
+        super
       end
     end
   end
--- a/config/workflows/one_step.json
+++ b/config/workflows/one_step.json
@@ -46,7 +46,7 @@
             "notifications": [{
                 "notification_type": "email",
                 "name": "ScholarsArchive::Workflow::DepositedNotification",
-                "to": ["approving"]
+                "to": ["depositing"]
             }],
             "methods": [
                 "Hyrax::Workflow::RevokeEditFromDepositor",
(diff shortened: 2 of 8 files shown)


app/services/scholars_archive/workflow/pending_review_notification.rb
```diff
@@ -11,8 +11,9 @@ def message
         "The item: #{title} (#{link_to work_id, document_path}) was deposited by #{user.user_key} and is awaiting review. #{comment}"
       end
 
+      # Add the user who initiated this action to the list of users being notified
```
revgum: Comment is invalid in this context I think

config/workflows/one_step.json
```diff
@@ -14,7 +14,7 @@
             }, {
                 "notification_type": "email",
                 "name": "ScholarsArchive::Workflow::DepositReceivedNotification",
-                "to": ["depositing"]
+                "to": []
```
revgum: Wouldn't you still want "to" users, or do these happen by default?

app/services/scholars_archive/workflow/pending_review_notification.rb
luisgreg99: comment removed

config/workflows/one_step.json
luisgreg99: I ended up reverting this change, refined the notification object instead

PR #762, opened 2017-07-25
Feature/default data

straleyb:


app/controllers/concerns/scholars_archive/works_controller_behavior.rb
```diff
@@ -7,6 +7,13 @@ module WorksControllerBehavior
       before_action :set_geo, only: [:create, :update]
     end
 
+    def new
+      curation_concern.publisher = ["Oregon State University"]
+      curation_concern.rights_statement = ["http://rightsstatements.org/vocab/InC/1.0/"]
+      curation_concern.degree_grantors = "http://id.loc.gov/authorities/names/n80017721"
```
luisgreg99: For `degree_grantors`, I think we should check if it's defined as well:
`curation_concern.degree_grantors = "http://id.loc.gov/authorities/names/n80017721" if curation_concern.respond_to?(:degree_grantors)`

PR #770, opened 2017-07-28
Fixes form width and Location styling

### LINKED ISSUE #668: Location dropdown on form needs to match style of other dropdowns
Currently, the location dropdown UI looks different from other dropdown UIs. Location dropdown UI should match other UI in Hyrax.

### LINKED ISSUE #680: make form fields all the same width (to fill the partial)
form fields are a bunch of different widths and they should all be consistent. 

Possibly push this change up to hyrax.

straleyb:
Fixes #680 
Fixes #668

### DIFF
--- a/app/assets/stylesheets/osu.scss
+++ b/app/assets/stylesheets/osu.scss
@@ -61,3 +61,36 @@ div#date_wrapper, div#geo_wrapper {
 .input-group-addon, .input-group-btn {
   width: 0% !important;
 }
+
+.select2-container {
+  width: 125% !important;
+  border: none !important;
+}
+
+.form-group .controlled_vocabulary {
+  width: 100% !important;
+}
+
+.select2-default {
+ background-color: none !important;
+}
+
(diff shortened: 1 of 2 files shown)


straleyb:
So I didnt want to keep spinning my wheels on this, but targeting the proper divs was kind of difficult without using the important tags. This dropdown uses a library called select2 which is a library made by Jquery and hosted through github. It is a javascript library which autogenerates html based on JS. This work was imported from scholarsphere into sufia a little while ago. One thing that I noticed about select2 is they tend to use important tags as well to get their style to stick. 

Select2 website >>>> https://select2.github.io/

With that being said, the style is mostly matching the other select drop downs with a little bit of work needing to be done to extend the length of the box and push the "remove" button to the right a bit more. Here is a screenshot of what I have. 

screen shot 2017-07-31 at 2 43 39 pm

vantuyls:
drop down definitely looks better. thanks!

PR #793, opened 2017-08-01
Feature/related url with label

### LINKED ISSUE #759: Add a label for Related Items field
In most cases, it will be helpful to end-users to know HOW the link is related to the item.

luisgreg99:
image

fixes #759 

- Added `NestedRelatedItems` with properties `label` (DC.title) and `related_url` (RFDS.seeAlso) for the backend. 
- Used `DC.relation` as the predicate for the parent `NestedRelatedItems`, but it can be changed if needed. 
- For the front-end, added `MultiValueLabelUrlInput` (inherits from `MultiValueInput` in hydra editor https://github.com/samvera/hydra-editor/blob/master/app/inputs/multi_value_input.rb)
- `MultiValueLabelUrlInput` was extended from `MultiValueInput` to render multiple inputs (one for each property in `NestedRelatedItems`) and support html classes, names and ids used in the form.


wickr:
I think 'Nested References' should be changed to 'Related Items' on the form and 'Nested Related Items' or equivalent in the code.

"References" carries weight more along the lines of 'works cited' to me.

Since we're not currently using DC.relation that may work for the predicate pointing to the parent nested object. @vantuyls ?

Also you should drop `:related_url` from default_metadata

app/assets/javascripts/nested_field.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+        removed.hide()
+        removed.appendTo($(this))
+    )
+    $(".multi_value.nested-field.form-group").on("managed_field:add", (event) ->
+      # Find the last child field and update the index (increment its value) in the name and id attributes so that they
+      # follow these patterns needed for a nested attribute:
+      #   name = "#{object_name}[#{attribute_name}_attributes][#{index}][label]"
+      #   id = "#{object_name}_#{attribute_name}_attributes_#{index.to_s}_label"
+      # example:
+      #   name = "article_nested_related_items_attributes][1][label]"
```
revgum: I think this comment should be `article[nested_related_items_attributes][1][label]` correct?

app/assets/javascripts/nested_field.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+      #   id = "article_nested_related_items_attributes_1_label"
+      window.setTimeout(->
+        target = $(event.target)
+        last_child = target.find(".field-wrapper").last()
+        last_child.find("div[class$='_id']").remove()
```
revgum: I feel like it would be nice to have a method to `removeLastChildId(last_child)` which finds and removes the last childs ID.

app/assets/javascripts/nested_field.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+        inputs = last_child.find('input')
+        firstChild = inputs.first()
+        matches_array = firstChild.attr("name").match(/\[(.*)\]\[(.*)\]\[(.*)\]/)
```
revgum: I think a method like to encapsulate the finding of the first child input of the last child, and getting its id would be nice.. then you would just call the method and use the id if it is greater than -1 .. 

```
function getLastChildId(last_child) {
  // find the inputs
  // find the first input
  // find the name matches
  // if name matches? {
  // -- parse int the name matches[2]
  // } else {
  // -- return -1 if the name matches[2] doesn't exist
  // }
}
```

app/assets/javascripts/nested_field.js.coffee
luisgreg99: Yes, that's correct, I have pushed an update with the change

PR #800, opened 2017-08-01
Fixes Default admin set issue and hidden form elements

### LINKED ISSUE #797: hide the relationship tab for non-admin users
Render the tab/partial, but add a class so that it will be `display:none` and hidden.. this way the form will post with the proper values but the tab isn't exposed to the user

### LINKED ISSUE #798: [Bug] Default Admin Set not being set properly
Array format is improper. Select from active_options instead.

### LINKED ISSUE #799: Hidden form element instead of non rendered for relationships tab

straleyb:
Fixes #797 
Fixes #798
Fixes #799

### DIFF
--- a/app/views/hyrax/base/_form_relationships.html.erb
+++ b/app/views/hyrax/base/_form_relationships.html.erb
@@ -1,10 +1,11 @@
-<% if Flipflop.assign_admin_set? %>
-  <%# TODO: consider `Hyrax::AdminSetOptionsPresenter.select_options_for(controller: controller)` instead %>
-  <%= f.input :admin_set_id, as: :select,
-      selected: ScholarsArchive::AdminSetSelectService.select(f.object.model.class.to_s, Hyrax::AdminSetOptionsPresenter.new(Hyrax::AdminSetService.new(controller)).select_options),
-      include_blank: true,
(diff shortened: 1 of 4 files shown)


app/views/hyrax/base/_form_relationships.html.erb
```diff
@@ -1,10 +1,11 @@
...
-  <%= f.input :admin_set_id, as: :select,
-      selected: ScholarsArchive::AdminSetSelectService.select(f.object.model.class.to_s, Hyrax::AdminSetOptionsPresenter.new(Hyrax::AdminSetService.new(controller)).select_options),
-      include_blank: true,
-      collection: Hyrax::AdminSetOptionsPresenter.new(Hyrax::AdminSetService.new(controller)).select_options,
-      input_html: { class: 'form-control' } %>
+<div class=<%= current_user.admin? ? "" : "hidden" %>>
+  <% if Flipflop.assign_admin_set? %>
+    <%# TODO: consider `Hyrax::AdminSetOptionsPresenter.select_options_for(controller: controller)` instead %>
+    <%= f.input :admin_set_id, as: :select,
+    selected: ScholarsArchive::AdminSetSelectService.select(f.object.model.class.to_s, Hyrax::AdminSetOptionsPresenter.new(Hyrax::AdminSetService.new(controller)).select_options),
```
revgum: I think the formatting of this `f.input` needs to be fixed with some left indent

PR #801, opened 2017-08-01
Fixes Work select modal to align and be 2 columns

### LINKED ISSUE #632: make work type pop up a grid instead of a list
with so many work types, changing the UI for this list would be helpful. make it a 2 by n grid. 

Happy to consult on what that could/should look like.

straleyb:
Fixes #632


straleyb:
screen shot 2017-08-02 at 8 51 09 am

PR #821, opened 2017-08-07
added calendar widget to date collected

### LINKED ISSUE #681: DateRange widget (ingest form, not date range facet)
for `date_collected` being a multi-value field, bind to the click event of adding a new date and bind the attachment of the widget to the new field that was created (or, any date element which doesn't already have the date range widget attached)

### LINKED ISSUE #785: [Date Collected] Date field should not have Date Collected already selected
when one enters an empty metadata form, the Date field shows Date Collected automatically. Would be preferable for no date to be pre-selected. If a date must be pre-selected it should be Date Copyright

### LINKED ISSUE #786: [Date Collected] Date Collected needs placeholder text in the date field box
all other dates show YYYY-mm-dd as help text in the date box. Date collected should do this, too.

### LINKED ISSUE #788: [Date Collected] Date Collected should have calendar widget and functionality like other date fields

luisgreg99:
fixes #788 
fixes #681 
fixes #786 
fixes #785

### DIFF
--- a/app/assets/javascripts/add_date_field.js.coffee
+++ b/app/assets/javascripts/add_date_field.js.coffee
@@ -1,7 +1,7 @@
 (($) ->
   Blacklight.onLoad ->
-    date_default = () ->
-      $('#date_wrapper .date-default .date-input input').datetimepicker({
+    date_default = (selector) ->
+      selector.datetimepicker({
         timepicker:false,
         format:'Y-m-d',
         scrollMonth: false,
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/add_date_field.js.coffee
```diff
@@ -164,4 +176,15 @@
 
       date_default()
       date_range(selected_date)
+
+    $(document).on("managed_field:add", '.multi_value_date.form-group', (event) ->
+      window.setTimeout(->
+        date_multiple_default()
```
revgum: Does `event.target` refer to the new managed field that was added? And if so, it would be a good idea to pass that to `date_multiple_default()` and call .datetimepicker directly on it instead of the current broad jquery selector for all of the fields that would have already had the .datetimepicker called on them.

app/assets/javascripts/add_date_field.js.coffee
```diff
@@ -10,7 +10,18 @@
         validateOnBlur: false
       })
 
+    date_multiple_default = () ->
```
revgum: Related to comment below, take `selector` as an argument to this method and call .datetimepicker on it.

Something like;

```
date_multiple_default = (selector) ->
  selector.datetimepicker({
...
```

app/assets/javascripts/add_date_field.js.coffee
```diff
@@ -10,7 +10,18 @@
...
+        timepicker:false,
+        format:'Y-m-d',
+        scrollMonth: false,
+        scrollTime: false,
+        scrollInput: false,
+        validateOnBlur: false
+      })
+
     date_default()
+    date_multiple_default()
```
revgum: Pass the full broad selector to bind datetimepicker to here `date_multiple_default($('#date_wrapper .date-multiple .date-input input'))`

PR #827, opened 2017-08-09
smart caching

straleyb:


app/services/scholars_archive/caching_service.rb
```diff
@@ -2,21 +2,68 @@
...
+      fetch_or_cache_json_from_etag(uri, expires_in)
     end
 
     private
+      def self.fetch_or_cache_json_from_etag(uri, expires_in)
+        etag = fetch_etag(uri)
+        data = read_etag_from_cache(uri)
+        json = nil
+        if data.nil? || data != etag
+          cache_etag(uri, etag, expires_in)
```
revgum: rather than catching the etag here, fetch the json, grab the etag from the fetched json, cache the json, and finally cache the etag from the fetch. 

The benefit here is that you're using the etag from the request that supplied the data, and only caching the etag after a successful fetching and caching of data.

PR #831, opened 2017-08-14
updated triple_powered_service to query rdf dc date

### LINKED ISSUE #706: Academic Affiliation & Other Affiliation label facet duplicates
Facets by TriplePoweredProperty related labels are showing as duplicates. This makes sense, because the `Music` Academic Unit is found in multiple URIs (because of date ranges likely) but leaves for a confusing experience. 

screen shot 2017-07-06 at 7 42 11 am

luisgreg99:
image

updated `ScholarsArchive::TriplePoweredService.fetch_all_labels` to append `RDF::Vocab::DC.date` info to fetched labels when available 

fixes #706

### DIFF
--- a/app/indexers/default_work_indexer.rb
+++ b/app/indexers/default_work_indexer.rb
@@ -18,10 +18,10 @@ class DefaultWorkIndexer < Hyrax::WorkIndexer
       language_labels = ScholarsArchive::LanguageService.new.all_labels(object.language)
       peerreviewed_label = ScholarsArchive::PeerreviewedService.new.all_labels(object.peerreviewed)
 
-      object.triple_powered_properties.each do |field|
-        labels = ScholarsArchive::TriplePoweredService.new.fetch_top_label(object.send(field))
-        solr_doc[field.to_s + '_label_ssim'] = labels
-        solr_doc[field.to_s + '_label_tesim'] = labels
+      object.triple_powered_properties.each do |o|
+        labels = ScholarsArchive::TriplePoweredService.new.fetch_top_label(object.send(o[:field]), parse_date: o[:has_date])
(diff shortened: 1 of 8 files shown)


app/indexers/default_work_indexer.rb
```diff
@@ -17,13 +17,19 @@ def generate_solr_document
       license_labels = ScholarsArchive::LicenseService.new.all_labels(object.license)
       language_labels = ScholarsArchive::LanguageService.new.all_labels(object.language)
       peerreviewed_label = ScholarsArchive::PeerreviewedService.new.all_labels(object.peerreviewed)
+      academic_affiliation_labels = ScholarsArchive::TriplePoweredService.new.fetch_top_label(object.academic_affiliation, parse_date:true)
```
revgum: Both of these calls are covered already by the `object.triple_powered_properties...` block below. 

Instead of these two calls, maybe update the block below to be like;
 ```
object.triple_powered_properties.each do |o|
  labels = ScholarsArchive::TriplePoweredService.new.fetch_top_label(object.send(o[:field]), o[:has_dates])
  solr_doc[o[:field].to_s + '_label_ssim'] = labels
  solr_doc[o[:field].to_s + '_label_tesim'] = labels
end

revgum: And then in has_triple_powered_properties.rb

```
      def triple_powered_properties
        [{field: :academic_affiliation, has_date: true}, {field: :other_affiliation, has_date: true}]
      end
```

revgum: ...and finally, there are a test or two that would need to respect this data structure as well as a has_url_validator.rb and triple_powered_form.rb

PR #846, opened 2017-08-22
Adds other embargo reason.

straleyb:
Adds in text field for "Other embargo reason"

### DIFF
--- a/app/assets/javascripts/embargo_reason_field.js
+++ b/app/assets/javascripts/embargo_reason_field.js
@@ -1,0 +1,15 @@
+$( document ).ready(function() {
+  $('#embargo_reason_select').change(function() {
+    var val = $("#embargo_reason_select option:selected").val();
+    if (val == 'Other...') {
+      $('#article_embargo_reason').removeClass('hidden');
+      $('#article_embargo_reason').val("");
+    } else {
+      $('#article_embargo_reason').addClass('hidden');
+      $('#article_embargo_reason').val(val);
+    }
+  });
+  $('input[type="radio"]#article_visibility_embargo').on('change', function(e) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/embargo_reason_field.js
```diff
@@ -0,0 +1,11 @@
+$( document ).ready(function() {
+  $('#embargo_reason_select').change(function() {
```
revgum: If the user leaves the default selection without triggering a "change" on the select, what would happen?

app/assets/javascripts/embargo_reason_field.js
```diff
@@ -0,0 +1,11 @@
...
+    var val = $("#embargo_reason_select option:selected").val();
+    if (val == 'Other...') {
+      $('#article_embargo_reason').removeClass('hidden');
```
revgum: clear the value for in the input, allowing the user to enter their own.

app/assets/javascripts/embargo_reason_field.js
```diff
@@ -0,0 +1,11 @@
...
+    } else {
+      $('#article_embargo_reason').addClass('hidden');
+      $('#article_embargo_reason').val(val);
+    }
+  });
```
revgum: I feel like when the embargo radio is selected, you would want to trigger "change" initially for the embargo_reason_select

app/assets/javascripts/embargo_reason_field.js
straleyb: So if they dont trigger a change, it accepts the form per usual. I didnt want to make the embargo reason required. What I can do though, if it sounds like a good idea, is to pre-fill the value for them and make not filling embargo reason impossible.

PR #851, opened 2017-08-23
Default query and filter in view

straleyb:
With just greg's fix, the lists weren't filtering. So I had to figure out some way to filter the lists out. This allows the lists to be filtered but in the view. The other place to filter this would be here. https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/admin/workflows_controller.rb#L17

However, @status_list is a status_list_service that has select as a private method. It makes it slightly difficult to filter out the proper objects. The only other way I could see it filtered is using the proper solr_query but that seems like a low level issue that we chose to not look into right now. For now, I propose just filtering in the view and calling it a day. Ill make a ticket that describes the issue and make comments where necessary.


straleyb:
https://github.com/samvera/hyrax/issues/1395 This is the hyrax issue with the updated information on the error that will eventually fix this issue.

Issue #595, opened 2017-05-24
Handle and URL Mapping/Redirects


vantuyls:
We need to respect links and handles from DSpace for SA@OSU on Hydra. Examples of URLs and Handles we need to respect are, e.g.:

* Item Handle: http://hdl.handle.net/1957/60832
* Item Record URL: http://ir.library.oregonstate.edu/xmlui/handle/1957/60832
* Item Record Full Showpage: http://ir.library.oregonstate.edu/xmlui/handle/1957/60832?show=full
* File Download URL: http://ir.library.oregonstate.edu/xmlui/bitstream/handle/1957/60832/main.pdf?sequence=1
* Community:    (redirect to either Admin Set or Solr search results with varying facets)
* Collection:    (redirect to either Admin Set or Solr search results with varying facets)
* any admin pages or other access points?

also see: #309, #274, #275, #596

@revgum @straleyb @zhang4952 @wickr @luisgreg99 

- [ ] On ir.l.o.e, redirect traffic to handle traffic to SA on a handle controller
- [ ] The handle controller derives the original handle uri and searches SOLR for this url (ie. http://hdl.handle.net/1957/89921)  
  - [ ] If a work is found, redirect user to `#show` to that work
  - [ ] If a work is not found, show a stylish 404 error and notify SA administration of the problem

revgum:
- [ ] On ir.l.o.e, redirect traffic to handle traffic to SA on a handle controller
- [ ] The handle controller derives the original handle uri and searches SOLR for this url (ie. http://hdl.handle.net/1957/89921)  
  - [ ] If a work is found, redirect user to `#show` to that work
  - [ ] If a work is not found, show a stylish 404 error and notify SA administration of the problem

straleyb:
1. Handle URI will be redirected to ir.l.o.e per usual. The uri route for ir.l.o.e will be in the form `/xmlui/bitstream/handle/1957/60832/main.pdf?sequence=1`
2. ir.l.o.e will be redirected to lib-saapp1.l.o.e with the uri route being the same. (RO JOB.) e.g. `ir.l.o.e/xmlui/bitstream/handle/1957/60832/main.pdf?sequence=1 => lib-saapp1.l.o.e/xmlui/bitstream/handle/1957/60832/main.pdf?sequence=1`
3. Controller in SA handles the request in the form `/xmlui/bitstream/handle/:handle_id/:id/:filename/(:sequence)`
[...]

PR #873, opened 2017-08-28
Handle Redirects

straleyb:


revgum:
@straleyb per ticket #595, I think it'd be cleaner to have a route and controller that would handle the urls expressed by the Item Record and File Download urls. This way the webserver only has to rewrite requests from ir.l.o.e to lib-saapp1.l.o.e but the rest of the request remains the same. This is why I suggested you put together a sort of "spec" describing how you'd handle the requests versus putting together code already.

revgum:
The "referrer" is more likely to be "ir.l.o.e" instead of the handle server.

straleyb:
Ok so if im understanding correctly I should only account for ir.l.o.e urls and not worry about the handle urls. I was pretty sure I put together a sufficient "spec" here for myself, but maybe im misunderstanding what you mean by spec. The spec is what informed this code that I wrote. I had most the comments in there before I even wrote the code and filled in as necessary.

straleyb:
On a second note, im not sure how I would handle setting up ir.l.o.e to redirect traffic to SA@OSU. Im assuming that would be a job for @rordway ?

revgum:
@straleyb yea the comments describe well what you're doing, but also under the assumption that the traffic is coming from the handle server. I was thinking a bit higher level description of the flow of the traffic to cover any/all bases.. for instance;

1. Traffic to the handle server gets redirected to ir.l.o.e (http://hdl.handle.net/1957/1234 redirects to http://ir.l.o.e/xmlui/handle/1957/1234)
1. Traffic to ir.l.o.e server needs to be redirected to lib-saapp1.l.o.e with the querystring intact (http://ir.l.o.e/xmlui/handle/1957/1234 redirected to http://lib-saapp1.l.o.e/xmlui/handle/1957/1234). _this is a task for RO_
1. A controller on SA should handle requests from the previous step (xmlui/handle/:handle_id/:id, xmlui/bitstream/:handle_id/:id)
1. The controller should interpret the handle based on the `handle_id` and `id` values configured by the route
1. ...
1. ...

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Hyrax
+  class HandlesController < ApplicationController
+    def handle_show
+      if params[:handle_id] != "1957"
+        puts "Handle ID does not match with Scholars Archive Handle ID."
+      elsif params[:id].length != 5
```
revgum: params[:id] isn't always 5..

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+        puts "ID is not the right length for Scholars Archive Handle ID."
+      end
+
+      solr_doc = query_solr_for_work(construct_handle_url(params[:id]))
+      work = query_fedora_for_work(solr_doc.first[:id], solr_doc.first[:human_readable_type_tesim].first.constantize)
+      redirect_to "/concern/#{work.class.to_s.downcase}s/#{work.id}"
```
revgum: I feel like you should use `pluralize` or look at the Rails doc to see if there is a "class to plural string" type of a method. Either way, adding `s` to the end of the class name isn't ever safe for trying to pluralize word.

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def handle_download
+    end
+
+    private
+      def construct_handle_url(handle_code)
+        "hdl.handle.net/1954/" + handle_code
```
revgum: since you have reference to it, you should take in and use the `handle_id` param.. avoid hardcoding

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      def query_fedora_for_work(id, work_type)
+        work_type.find(id)
+      end
+
+      def query_solr_for_work(handle)
+        #Query solr.
+        ActiveFedora::SolrService.query("replaces_tesim:#{handle}", :rows => 1000000)
```
revgum: does `replaces_tesim` have the full "http://hdl.handle.net...." or just the portion you're passing in as `handle`.. AFAICT it should include the `http://`

config/routes.rb
```diff
@@ -26,6 +27,12 @@
   curation_concerns_basic_routes
   concern :exportable, Blacklight::Routes::Exportable.new
 
+  Hyrax::Engine.routes.draw do 
+    get '/xmlui/handle/:handle_id/:id', to: 'handles#handle_show', as: 'handle_basic_show'
+    get '/xmlui/handle/:handle_id/:id/(:show)', to: 'handles#handle_show', as: 'handle_specific_show'
```
revgum: routes with optional params (like you saw on the example urls, no doubt) aren't necessary. I think, in general, we'd just want to redirect people to the show page of the work that download could be found in (or the show page for a regular link)

You don't need specify the querystring params: http://guides.rubyonrails.org/routing.html#the-query-string

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,80 @@
+module Hyrax
+  class HandlesController < ApplicationController
```
revgum: I think you need a before action instead of the "if checks" on your methods.. you don't want the methods to proceed if handle_id is invalid and this isolates your validation checking to a single simple place.
```
before_action :validate_handle_id, only: [:handle_show, :handle_download]
...

private 
def validate_handle_id
  if params[:handle_id] != "1957"
    puts ..
    redirect_to '/404/file_not_found'
  end
end
```

revgum: ..and use your logging service, so we can get notification about this sort of traffic?

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+        filesets.flatten.compact
+      end
+
+      def find_work
+        solr_doc = query_solr_for_work(construct_handle_url(params[:handle_id], params[:id]))
+        query_fedora_for_work(solr_doc.first[:id], solr_doc.first[:human_readable_type_tesim].first.constantize)
+      end
+
+      def construct_handle_url(handle_id, handle_code)
+        "hdl.handle.net/#{handle_id}/" + handle_code
```
revgum: stick with consistent string concatenation

`"hdl.handle.net/#{handle_id}/#{handle_code}"`

app/services/scholars_archive/handle_error_logging_service.rb
```diff
@@ -0,0 +1,20 @@
+module ScholarsArchive
+  class HandleErrorLoggingService
+    def self.log_no_files_found_error(handle_params, work, handle_url)
```
revgum: I feel like we should add a ticket related to this service to fire off an email to `ENV['SCHOLARSARCHIVE_ADMIN_EMAIL']` (or whatever the new ENV Greg just created) for these errors as well.

app/controllers/hyrax/handles_controller.rb
```diff
@@ -0,0 +1,80 @@
+module Hyrax
+  class HandlesController < ApplicationController
+    def handle_404
+
```
revgum: This view could have a nice message about the traffic failing?

app/views/scholars_archive/handles/handle_file_404.html.erb
```diff
@@ -0,0 +1,2 @@
+<p> The work you are trying to look at <b><%= @handle_uri %></b> doesnt seem to have the file you requested <b><%= @file  %></b>. If you believe this is an error please notify an administrator. Thank you.</p>
```
revgum: spelling error on `doesnt`

app/services/scholars_archive/handle_error_logging_service.rb
straleyb: https://github.com/osulp/Scholars-Archive/issues/881

app/controllers/scholars_archive/handles_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+      elsif filesets.length > 1
+        ScholarsArchive::HandleErrorLoggingService.log_too_many_files_found_error(params, work, construct_handle_url(params[:handle_id], params[:id]))
+        redirect_to "/concern/#{work.class.to_s.downcase.pluralize}/#{work.id}"
+      else
+        redirect_to "/downloads/#{filesets.first.id}"
+      end
+    end
+
+    private
+      def verify_handle_id
```
wickr: To me `:handle_id` means the whole thing, and using it only for the prefix is confusing. From the Handle docs: `Handles consist of a prefix which identifies a "naming authority" and a suffix which gives the "local name" of a resource.`.
I would suggest changing all instances (including methods and log output):
* change `:handle_id` to `:handle_prefix`
* change `:id` to `:handle_localname`

app/controllers/scholars_archive/handles_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    def handle_work_404
+
+    end
+
+    def handle_show
+      work = find_work
+      if work.nil?
+        ScholarsArchive::HandleErrorLoggingService.log_no_work_found_error(params)
+        redirect_to "/404/work_not_found"
```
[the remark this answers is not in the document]
wickr: Some of the options in there may allow you to use more of the built-in Rails stuff and not need custom 404 routes and views just to display the extra information.

app/views/scholars_archive/handles/handle_file_404.html.erb
```diff
@@ -0,0 +1,2 @@
+<p> The work you are trying to look at <b><%= @handle_uri %></b> doesn't seem to have the file you requested <b><%= @file  %></b>. If you believe this is an error please notify an administrator. Thank you.</p>
```
wickr: I'd add a note that this error was logged. Then something like "If you believe this is an error, you can use the Contact Form to let us know."

app/views/scholars_archive/handles/handle_file_404.html.erb
```diff
@@ -0,0 +1,2 @@
...
+<p> You can find the works main page here: <a href=<%= @work_path %>> Link to Work </a>
```
wickr: 'works' is missing an apostrophe. Instead of 'Link to Work' maybe display the title of the work? I'm thinking people may not understand some of these terms. So simplifying it to something like "The file requested may be available as part of [Title of Parent Work]"

app/controllers/scholars_archive/handles_controller.rb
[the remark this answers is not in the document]
straleyb: In order to have multiple different 404 pages with useful information, we definitely need to make some custom routes for the 404. Rails only knows about the /404 route, so no matter what we would need to make at least one new route. I figured we could leave the original 404 page as needed if someone goes anywhere else in the app that returns a 404. Otherwise, when they get the 404 messages from the handle redirect, they get some useful info and can make a decision after that.

I can change it so it renders a page with the proper status code for sure though.

app/controllers/scholars_archive/handles_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+        end
+        filesets.flatten.compact
+      end
+
+      def find_work
+        solr_doc = query_solr_for_work(construct_handle_url(params[:handle_prefix], params[:handle_localname]))
+        query_fedora_for_work(solr_doc.first[:id], solr_doc.first[:human_readable_type_tesim].first.constantize) unless solr_doc.empty?
+      end
+
+      def construct_handle_url(handle_prefix, handle_code)
```
revgum: for consistency sake, you should rename `handle_code` to `handle_localname`?

PR #980, opened 2017-11-01
[WIP] updated robots.txt

luisgreg99:
I'm excluding the paths to oai, the catalog, and the file stats, like University of Alberta for this commit. Stanford seems to be excluding everything, I'm wondering if we should do that instead. Princeton only has a `Crawl-delay: 8` directive that seems to be for yahoo and bing. Should we also have this directive too?

Any suggestions @wickr @revgum @rordway?

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -3,3 +3,7 @@
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-agent: *
 # Disallow: /
+User-agent: *
+Disallow: /catalog
+Disallow: /files/*/stats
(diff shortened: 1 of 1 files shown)


luisgreg99:
It looks like Google ignores the crawl rate directive (see https://support.google.com/webmasters/answer/48620?hl=en&vid=0-241580346937-1509736169344) However, Bing recommends it only for sites with large amount of pages with integer values less than 10, see https://blogs.bing.com/webmaster/2009/08/10/crawl-delay-and-the-bing-crawler-msnbot

PR #1043, opened 2017-11-21
replace collections yml with yml that include communities and collections redirect

### LINKED ISSUE #1015: allow DSpace Collection URIs to resolve to SOLR query in Hyrax
use the DSpace_Collection metadata field to fire off a query for items in that DSpace_Collection. Redirect DSpace Collection handles to those queries.

zhang4952:
fix #1015


config/handles_ir_communities_collections.yml
```diff
@@ -0,0 +1,1145 @@
+handle_ir_communities_collections
```
luisgreg99: Replace line (1) with `handles_ir_communities_collections`. Then, update https://github.com/osulp/Scholars-Archive/blob/master/app/controllers/scholars_archive/handles_controller.rb#L51 and https://github.com/osulp/Scholars-Archive/blob/master/app/controllers/scholars_archive/handles_controller.rb#L62

config/handles_ir_communities_collections.yml
```diff
@@ -0,0 +1,1145 @@
...
+# 10th IIFET Conference, July 10-14, 2000, Corvallis, Oregon, U.S.A.  
+ "30554" : "https://ir.library.oregonstate.edu/catalog?q=10th%20IIFET%20Conference%2C%20July%2010-14%2C%202000%2C%20Corvallis%2C%20Oregon%2C%20U.S.A.&search_field=dspace_community"
```
luisgreg99: This line should be:
`"30554": "https://ir.library.oregonstate.edu/catalog?q=10th%20IIFET%20Conference%2C%20July%2010-14%2C%202000%2C%20Corvallis%2C%20Oregon%2C%20U.S.A.&search_field=dspace_community"`

PR #1122, opened 2017-12-16
[WIP] added task to set conference name based on known dspace collection

### LINKED ISSUE #1096: Only 1 value in Conference Name facet
wickr commented 21 days ago

There should probably be more values here. Was the data not in DSpace to start with? Did the data get migrated to a wrong or incorrect predicate?

If we need to pull data out of other field(s) then this may be a project Sarah can help with.
@wickr wickr added Post-Migration Work question labels 21 days ago
@vantuyls
Member
vantuyls commented 18 days ago

i suspect this is erroneous mapping. I'll talk to @simholt tomorrow about this.

@zhang4952 is there any way you can think of to identify conference proceedings in DSpace?
@wickr
Owner
wickr commented 17 days ago

@vantuyls @boockm didn't think the data was in the records as a separate field, but it was in the Collection name. He added 'conference name' column and values to this spreadsheet: https://docs.google.com/spreadsheets/d/1Z_G-LO65NYf5BqkMH0koLzG5DoZ8KD3yAapViWCxiJw/edit#gid=0

So we can probably do this in a batch update.
@wickr wickr referenced this issue in osulp/Scholars-Archive 9 days ago
Open
Add conference field metadata #1055

luisgreg99:
fixes #1096

### DIFF
--- a/lib/tasks/set_conference_name.rake
+++ b/lib/tasks/set_conference_name.rake
@@ -78,7 +78,7 @@ namespace :scholars_archive do
 
     # perform updates
     collection_conf_name_mapping.each do |collection, conf_name|
-      solr_query_str = "dspace_collection_tesim:\"#{conf_name}\""
+      solr_query_str = "dspace_collection_tesim:\"#{collection}\""
       docs = ActiveFedora::SolrService.query(solr_query_str, {:rows => 100000})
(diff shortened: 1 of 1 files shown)


wickr:
Looks like the mapping of Collections got off somehow.

luisgreg99:
@wickr I've updated the mapping with the correct values, thanks for letting me know

Issue #1141, opened 2017-12-21
DSpace Community Handles redirects missing


wickr:
DSpace Communities with sub-Communities were not included in the Handle redirect list, and redirects are broken.

Since the parent Community title won't be in any item metadata, we'll need to figure out what to link to. Could do an OR'd search for the sub-Communities title(s).

- [x] 2870: Cascades Campus [21 items]: 
  * Community: "Faculty Research" OR "Student Research" 
  * Collection: 
    * "Faculty Research Publications (Cascades)" [4 items] OR 
    * "Presentations and Posters (Cascades)" [1 item] OR 
    * "Cascades Poster Session" [15 items] OR 
    * "Theses (Cascades)" [1 item]
  * Use: https://ir.library.oregonstate.edu/catalog?utf8=%E2%9C%93&search_field=dspace_collection&q=%5C+%28%22Faculty+Research+Publications+Cascades%22+OR+%22Presentations+and+Posters+Cascades%22+OR+%22Cascades+Poster+Session%22+%22Theses+Cascades%22%29  [reports 19 items, 2 are missing Collections due to #1165 ]

- [x] 1718: Department of Geosciences: 
  * Community: "Geography" OR "Geology" OR "Program in Water Conflict Management and Transformation"
  * Use: https://ir.library.oregonstate.edu/catalog?utf8=%E2%9C%93&search_field=dspace_community&q=%5C+%28%22Geology%22+OR+%22Geography%22+OR+%22Program%20in%20Water%20Conflict%20Management%20and%20Transformation%22%29

- [x] 9118: Provost's Office: 
  * Community: "Information Services" OR "University Information Technology Committee"
  * Use: https://ir.library.oregonstate.edu/catalog?utf8=%E2%9C%93&search_field=dspace_community&q=%5C+%28%22Information+Services%22+OR+%22University+Information+Technology+Committee%22%29

- [x] 59892: College of Science: 
  * Collection: "Presentations and Posters (College of Science)"
  * Use: https://ir.library.oregonstate.edu/catalog?q=%22Presentations%20and%20Posters%20(College%20of%20Science)%22&search_field=dspace_collection
- [x] Change 59893 to Collection: "Presentations and Posters (College of Science)"  (current 'Faculty Research' is used more than once): https://ir.library.oregonstate.edu/catalog?q=%22Presentations%20and%20Posters%20(College%20of%20Science)%22&search_field=dspace_collection
[...]

PR #1142, opened 2017-12-21
Add more missing Handles.

wickr:
Add more missing Collections Handles (and 1 Community Handle)


wickr:
Added more Community Handles, changed 'Faculty Research' for College of Science so it would just use the child Collection and give the correct results.
Addresses most, but not all, of the Handles in #1141

PR #1149, opened 2017-12-29
Fix spelling

wickr:
Fix degree name spelling that @simholt caught


simholt:
Terrible name, but it is a thing!

http://gradschool.oregonstate.edu/environmental-sciences/professional-science-master-degree-environmental-science

PR #1176, opened 2018-01-22
Includes recaptcha for the contact page

### LINKED ISSUE #378: Recaptcha on contact form
To help mitigate the amount of bots sending emails, we would like to add a recaptcha to the form and have it validate that someone using the form is a human being. 

https://github.com/ambethia/recaptcha
https://www.google.com/recaptcha/intro/index.html

straleyb:
Fixes #378 
This includes the recaptcha v2 from google via a rails gem.

The rails gem manages all the javascripts and makes it extremely easy to configure and include keys via a rails esque way of programming. 

This is my proposed fix for SA to get it in here. Changing this for inclusion in Hyrax could be a pain in the butt so I figure this would probably be best to push up now. 

My proposed fix for hyrax would be to set this up with flip flop to turn this captcha on and off but including it this way, using the rails gem might not be the way they want to go. Id much rather propose this fix to them and if it gets accepted in, then we can remove this code from SA and use hyrax to manage it instead.


app/views/hyrax/contact_form/new.html.erb
```diff
@@ -0,0 +1,52 @@
+<div class="alert alert-info">
+  <%= render 'directions' %>
+</div>
+
+<h1>
+    <%= t('hyrax.contact_form.header') %>
+</h1>
+
+<% if user_signed_in? %>
```
revgum: Variables set here aren't very clear.. can you just move them to where they are used directly 

`value: <%= user_signed_in? ? current_user.name : '' %>`
`value: <%= user_signed_in? ? current_user.email : '' %>`

config/local_env.example.yml
```diff
@@ -34,3 +34,7 @@ SCHOLARSARCHIVE_ADMIN_EMAIL: 'noreply@FQDN'
 
 # The full path to the directory that BrowseEverything uses for uploads, the webserver user must have access to this directory
 BROWSEEVERYTHING_FILESYSTEM_PATH: '/tmp/browse-everything'
+
+#Recaptcha configuration
+SCHOLARSARCHIVE_RECAPTCHA_SECRET_KEY: 'asdfasdfasdfasdfasdfadsfasdfadfsfa'
```
revgum: Don't forget to add these keys to Staging and Production when this is merged.

app/views/hyrax/contact_form/new.html.erb
straleyb: for sure. This is just how the code was when I copied it over from hyrax.

config/local_env.example.yml
straleyb: 👍 Will do for sure. I have it already set up along with BO's recaptcha stuff, so Im locked and loaded to get those in.

straleyb:
Ok working on tests now.

PR #1188, opened 2018-01-26
reworks external/search linking and consistent styling

### LINKED ISSUE #1170: New behavior for external and internal links
the following spreadsheet should be used to re-define how internal and external links behave from the work showpage:

https://docs.google.com/spreadsheets/d/1rDBHRrKqsIuVZuSrjWNTO-CVeBvJURC924M_4pipKqU/edit#gid=0

revgum:
fixes #1170 

search-external-links


app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb
```diff
@@ -15,16 +15,52 @@ def search_field
...
+      def maybe_license_uri(term)
+        puts term
+        yaml = YAML.load_file(File.join(Rails.root, 'config', 'authorities', 'licenses.yml'))
+        license = yaml['terms'].find { |l| l['term'].casecmp(term).zero? }
+        return '' if license.nil?
+        license['id']
+      end
+
+      def maybe_rights_statement_uri(term)
+        puts term
```
straleyb: Why is this puts here?

app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb
```diff
@@ -15,16 +15,52 @@ def search_field
...
+      def hash?(s)
+        s.include? '=>'
+      end
+
+      def maybe_uri(s)
+        URI.extract(s, %w[http https]).first || ''
+      end
+
+      def maybe_license_uri(term)
+        puts term
```
straleyb: puts?

straleyb:
Only thing I can see here is a bit of repeated code in `#maybe_license_uri` and `#maybe_rights_statement_uri` maybe something like

```
def extract_value_from_yaml(yaml)
 value = yaml['terms'].find { |l| l['term'].casecmp(term).zero? }
 return '' if value.nil?
 value['id']
end
```

PR #1196, opened 2018-01-30
Public default visibility

### LINKED ISSUE #1131: Set default visibility selection to nothing selected
though if there are particular challenges to that, setting the default to Public would also be fine.

revgum:
fixes #1131


straleyb:
Does this have any way of allowing things to be set to private? Or am I confusing what this does?

PR #1204, opened 2018-01-31
Added default values to model

### LINKED ISSUE #1174: Customize worktypes w/defaults
Resource Type: Honors College Thesis
NA Affiliation: Honors College

wickr:
Fixes #1174

### DIFF
--- a/spec/features/create_honors_college_thesis_spec.rb
+++ b/spec/features/create_honors_college_thesis_spec.rb
@@ -7,7 +7,6 @@ include Warden::Test::Helpers
 RSpec.feature 'Create a Honors College Thesis', js: false do
   context 'a logged in user' do
     let(:user) { User.first }
-    let(:role) { Role.find_or_create_by(name: "admin")}
     let(:current_user) { user }
 
     let(:admin_set) do
@@ -61,15 +60,11 @@ RSpec.feature 'Create a Honors College Thesis', js: false do
       Devise.cas_create_user = true
       User.authenticate_with_cas_ticket(@ticket)
 
-      allow(user).to receive(:admin?).and_return(true)
-      role.users << user
-
       login_as user
       visit new_hyrax_honors_college_thesis_path
     end
 
-    it "creates a new work" do
-      expect(page).to have_content "Add New Honors College Thesis"
+    it "renders the new form" do
       fill_in 'honors_college_thesis_title', with: 'Test Honors College Thesis'
       fill_in 'Creator', with: 'Test Honors College Thesis Creator'
       fill_in 'Commencement Year', with: '2018'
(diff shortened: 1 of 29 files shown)


spec/features/create_honors_college_thesis_spec.rb
```diff
@@ -24,22 +25,43 @@
...
+
     before do
       Hyrax::PermissionTemplateAccess.create(permission_template: permission_template, agent_type: 'user', agent_id: user.user_key, access: 'deposit')
       Sipity::WorkflowAction.create(id: 4, name: 'show', workflow_id: workflow.id)
+      allow_any_instance_of(ApplicationHelper).to receive(:select_tag_dates).and_return("")
+      allow_any_instance_of(Hyrax::DefaultForm).to receive(:date_terms).and_return([])
+      allow_any_instance_of(ScholarsArchive::AcademicUnitsService).to receive(:select_sorted_all_options).and_return(academic_unit_sorted_all_options)
+      allow_any_instance_of(ScholarsArchive::AcademicUnitsService).to receive(:select_sorted_current_options).and_return(academic_unit_sorted_all_options)
+      allow_any_instance_of(ScholarsArchive::DegreeFieldService).to receive(:select_sorted_all_options).and_return(degree_field_sorted_all_options)
+      allow_any_instance_of(ScholarsArchive::DegreeFieldService).to receive(:select_sorted_current_options).and_return(degree_field_sorted_all_options)
```
luisgreg99: I think we can add 
```
ENV[“SCHOLARSARCHIVE_DEFAULT_ADMIN_SET”] = 'Default Admin Set'
```
Here to set the env for the default admin set variable and pass the nil string check in circle.

spec/features/create_honors_college_thesis_spec.rb
```diff
@@ -24,22 +25,43 @@
...
+#      allow(controller).to receive(:current_user).and_return(current_user)
+
       login_as user
```
luisgreg99: I think also `login_as user` might not work as expected since we are using CAS authentication. I wonder it would work having  instead `RubyCAS::Filter.fake(user.onid)` like in https://github.com/osulp/Room-Reservation/blob/master/spec/features/admin/room_spec.rb#L8

revgum:
@luisgreg99 great updates, however I think we should move these default values out of the controllers and mix them into our models and the default metadata module.. consider this;

## app/models/concerns/scholars_archive/default_metadata.rb
```
...
    included do
      # Set the visibility for all works
      after_initialize :set_default_visibility
      def set_default_visibility
        self.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC if new_record?
      end
      # After setting default visibility, provide each model a hook to set property defaults
      after_initialize :set_defaults, unless: :persisted?
...
```

## app/models/conference_proceedings_or_journals.rb
```
...
  private
  def set_defaults
    self.peerreviewed ||= 'FALSE'
  end
```

PR #16, opened 2015-03-16
Restrict Ingest to Administrators

terrellt:
Closes #6

PR #1242, opened 2018-02-16
Adds PER to SA

straleyb:
Image not saving for some odd reason. Also, has to account for a nil value in the selector for the modal that displays the models that can be created in the system.


app/helpers/application_helper.rb
```diff
@@ -96,7 +96,7 @@ def facet_desc_sort!(items=[])
   def fixed_work_type_order(items=[])
     model_list = []
     items.each {|item| model_list.push(item) }
-    fixed_ordered_list = [Article, EescPublication, GraduateThesisOrDissertation, TechnicalReport, GraduateProject, AdministrativeReportOrPublication, HonorsCollegeThesis, ConferenceProceedingsOrJournal, UndergraduateThesisOrProject, OpenEducationalResource, Dataset, Default]
+    fixed_ordered_list = [Article, EescPublication, GraduateThesisOrDissertation, TechnicalReport, GraduateProject, AdministrativeReportOrPublication, HonorsCollegeThesis, ConferenceProceedingsOrJournal, UndergraduateThesisOrProject, OpenEducationalResource, Dataset, Default, nil, PurchasedEResource]
```
revgum: what purpose is the `nil` here for?

straleyb: Steve specifically wanted it to show up under default in the last row. So we had to skip a cell in the modal view.

straleyb: screen shot 2018-02-16 at 3 38 07 pm

straleyb:
```
vantuyls [2:11 PM]
bottom of the right side
keep two columns
if this creates a blank space at bottom left, thats fine
```

straleyb:
Simple objects for application_helper {visible: true, model: SomeModelHere}...

app/views/shared/_select_work_type_modal.html.erb
```diff
@@ -11,34 +11,38 @@
...
                       <label>
-                        <input type="radio" name="payload_concern" value="<%= row_presenter[0].concern %>"
-                               data-single="<%= new_polymorphic_path([main_app, row_presenter[0].concern]) %>"
-                               data-batch="<%= hyrax.new_batch_upload_path(payload_concern: row_presenter[0].concern) %>">
+                        <input type="radio" name="payload_concern" value="<%= row_presenter[0]["model"].concern %>"
+                               data-single="<%= new_polymorphic_path([main_app, row_presenter[0]["model"].concern]) %>"
+                               data-batch="<%= hyrax.new_batch_upload_path(payload_concern: row_presenter[0]["model"].concern) %>">
                         <div class="select-work-icon">
-                          <span class="<%= row_presenter[0].icon_class %>"></span>
+                          <span class="<%= row_presenter[0]["model"].icon_class unless row_presenter[0].nil? %>"></span>
```
revgum: I'd combine the nil? check alongside the visibility check on line #16 and get rid of these hanging `unless row_presenter[0].nil?`.

Also.. it might be worthwhile making a _select_work_type_cell.html.erb partial with lines 15 through 30, passing in each `row_presenter`.

PR #1257, opened 2018-02-27
Adds in smarter validation for unrequired fields

straleyb:
This is my proposition for fixing the unrequired issue for degree grantors. It allows for the extension of unrequired fields beyond just PurchasedEResources. If a record has this behavior, and responds to the method, then it'll make decisions based on whether the field is nil or empty. If so, it'll skip it, if it has a value, then it'll validate that value is correct.


lib/scholars_archive/triple_powered_properties/has_url_validator.rb
```diff
@@ -11,6 +11,7 @@ def validate(record)
       return if record.triple_powered_properties.empty?
 
       record.triple_powered_properties.each do |prop|
+        return if prop[:skip_validation] && record[prop].empty?
```
straleyb: So we have to check if also the field is empty because if there is a value added to the field then we definitely want to validate. So we will skip iff the skip_validation is set to true and the field is empty.

PR #1260, opened 2018-02-28
Fixes grad year validator to work with nil

### LINKED ISSUE #1255: graduation_year validation doesn't account for `Nil`
https://github.com/osulp/Scholars-Archive/blob/master/lib/scholars_archive/validators/graduation_year_validator.rb#L4

`return if record.graduation_year.present? && valid_value(record)`

straleyb:
Fixes #1255

### DIFF
--- a/lib/scholars_archive/validators/graduation_year_validator.rb
+++ b/lib/scholars_archive/validators/graduation_year_validator.rb
@@ -1,7 +1,7 @@
 module ScholarsArchive::Validators
   class GraduationYearValidator < ActiveModel::Validator
     def validate(record)
-      return if record.graduation_year.empty? || valid_value(record)
+      return if record.graduation_year.present? || valid_value(record)
       add_error_message(record)
       return
     end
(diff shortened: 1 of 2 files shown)


lib/scholars_archive/validators/graduation_year_validator.rb
```diff
@@ -1,7 +1,7 @@
 module ScholarsArchive::Validators
   class GraduationYearValidator < ActiveModel::Validator
     def validate(record)
-      return if record.graduation_year.empty? || valid_value(record)
+      return if record.graduation_year.present? || valid_value(record)
```
luisgreg99: I think this should be `!record.graduation_year.present?` instead of `record.graduation_year.present?`.

straleyb: I think you are right. I went off of what was in this ticket
https://github.com/osulp/Scholars-Archive/issues/1255

Issue #1311, opened 2018-04-10
link text color


revgum:
Link text color is black, which makes it unclear what links are on the system. This might make sense when looking at the values of metadata on the `show` page for a work, but definitely doesn't make sense everywhere on the application.

vantuyls:
need to look for an OSU web color that would work for link text.

vantuyls:
http://communications.oregonstate.edu/brand-guide/visual-identity/colors

Looking at Stratosphere, not sure i love this color for link text. @revgum @simholt @wickr @straleyb @KennaW @zhang4952 @luisgreg99 thoughts?

revgum:
HighTide looks awesome

revgum:
https://ir.library.oregonstate.edu/admin/appearance

I set the link and footer link color to high tide, give it a spin and lets talk about if this is good or if we need to reconsider.

@vantuyls

vantuyls:
looks good. thank!

simholt:
@revgum Isn't this already running on production?

revgum:
@simholt indeed, this isn't really a code change, just an update on the appearance settings

PR #1312, opened 2018-04-10
sets link text to an identifiable color for the help text

revgum:
ref #1311 (not fully addressed throughout the application)

This sets the link color, for the work type modal help text, to osu orange until #1311 is solved throughout the application.


simholt:
Nice links.

PR #1318, opened 2018-04-10
adds missing degree grantors

### LINKED ISSUE #1313: Make all degree grantors visible to admins
Right now, edit options in drop-down menu for Degree Grantor are Oregon State University and Other. Would like admins to see all versions of OSU, such as Oregon Agricultural College, Oregon State College, etc.

revgum:
fixes #1313 

screen shot 2018-04-10 at 2 37 34 pm


revgum:
TODO: The change to `config/initializers/hyrax.rb` must be reflected on Staging and Production in conjunction with the deployment that includes this code.

PR #1326, opened 2018-04-12
Adds in buttons for selecting pages dynamically

straleyb:
Looking for feedback. Refactoring I can see is creating a partial for the `<li><a>` elements cutting down on code duplication.


straleyb:
paginate

PR #1371, opened 2018-04-27
Feature/advanced search

### LINKED ISSUE #888: Allow for multiple facets on a single facet field [Advanced Search]
currently, when faceting on a field, one can only use a single facet per metadata field. often, though, we want to use multiple facets for a single field (e.g. i want to see all of the items by Steve AND Hui).

### LINKED ISSUE #1280: Enable Blacklight Advanced Search
some sorting and searching may be more functional with Blacklight advanced search enabled. This issue is to ask the question: is this a good idea, and, if yes, to initiate the investigation of what it would mean to implement this on SA@OSU. 

related: nothing

wickr:
Sets up Advanced Search with some customization to facets and search fields.

Fixes #1280 
Fixes #888

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,8 @@ gem 'blacklight_oai_provider', git: 'https://github.com/UNC-Libraries/blacklight
 gem 'blacklight_range_limit'
 gem 'edtf', github: 'inukshuk/edtf-ruby', branch: 'master'
 
+gem 'blacklight_advanced_search'
+
 # MySQL for staging/production servers
 gem 'mysql2', '~> 0.3.18'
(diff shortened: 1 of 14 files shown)


revgum:
For the most part, the problem was in the view override passing in `search_field` to the advanced_search_path.. this caused the blacklight_advanced_search form view to improperly inject a duplicated hidden_input without the proper value. 

Simple fix:
`  <%= link_to 'Advanced Search', blacklight_advanced_search_engine.advanced_search_path(search_state.to_h.except(:search_field)), class: 'advanced_search' %>`

straleyb:
advanced-search

PR #1398, opened 2018-05-14
remove identifier from forms and expose for admin

### LINKED ISSUE #1396: Hide Identifier field on ingest forms
We currently do not have a role for this field and people populate it with whatever random information they think is appropriate. Hiding it will save time and frustration for those who approve uploads and (hopefully) encourage users to put their info in more appropriate fields.

straleyb:
Fixes #1396


straleyb:
I still need to display screenshots, but im running into a weird error with Psych 
```
did not find expected key while parsing a block mapping at line 3 column 1 (Psych::SyntaxError)
```
When I try to boot the rails server. Any ideas? @revgum @luisgreg99

straleyb:
Identifier now shows up at the bottom as an admin
screen shot 2018-05-14 at 3 29 41 pm

No identifier at the end if non-admin
screen shot 2018-05-14 at 3 30 41 pm

PR #1415, opened 2018-05-21
Feature/display first label

### LINKED ISSUE #1288: Unwanted variant labels showing up for non-aca other
@wickr said this happened before and was fixed.

Created a new non-aca other:

> 
> - id: "http://id.loc.gov/authorities/names/n79100852"
> term: "Oregon Sea Grant"

When assigned, it lists **ALL** the variants from LC:

not all the variants

See here:
https://ir.library.oregonstate.edu/concern/defaults/1g05fh80z

straleyb:
Fixes #1288 

This Indexes the first label as the primary label and all others as the alt labels. If only one label is returned, it will index an empty array in the alt label field.

### DIFF
--- a/app/indexers/default_work_indexer.rb
+++ b/app/indexers/default_work_indexer.rb
@@ -21,10 +21,9 @@ class DefaultWorkIndexer < Hyrax::WorkIndexer
         uris = Array(object.send(o[:field]))
         uris = Array(object.send(o[:field])).reject { |u| u == "Other" }
         labels = ScholarsArchive::TriplePoweredService.new.fetch_top_label(uris, parse_date: o[:has_date])
-        solr_doc[o[:field].to_s + '_label_ssim'] = [labels.first]
-        solr_doc[o[:field].to_s + '_label_tesim'] = [labels.first]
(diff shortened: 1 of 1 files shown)


wickr:
Are we sure that prefLabel is always going to be first?  If the code fetches all labels, but it's returned as:  `[alt1, alt2, pref, alt3]` then just taking the first one won't be the preferred label and won't accurate.

wickr:
Ok, so `fetch_all_labels` returns them in order of the `rdf_label_predicates` found: https://github.com/osulp/Scholars-Archive/blob/master/app/services/scholars_archive/triple_powered_service.rb
Since prefLabel is first, you should be able to rely on `[pref, alt1, alt2, alt3]`. 

But `fetch_top_label` already does pick only the first one, it's just not called for fields with multiple set to true, in the indexer. I would look at tweaking that code, maybe if there's multiple values, iterate through those and call `fetch_top_label`.

Also I wouldn't add the `alt_label` indexing yet until we figure out how/where we want to do that.

straleyb:
So are you suggesting that we dont even check for multiples anymore? Just fetch the top label and call it good? I can make the changes for that.

PR #1418, opened 2018-05-22
Feature/update paging logic

### LINKED ISSUE #1153: Improve SA sorting by facet --> more
SA sorting by Facet-->More is A-Z or Numerical. That’s fine but having only the "Previous/Next" buttons is sooo slow, defeats purpose of sorting unless have few results.

Need other options: End, list page numbers, skip between pages, etc. 

screen shot 2018-01-04 at 12 59 53 pm

### LINKED ISSUE #1349: Catalog view, facet pagination
It appears that `numDocs` returns the full count of works having the facet being viewed.. but the visibility of the works aren't reflected in that number. Private, or otherwise hidden works, can cause the pagination to render multiple "pages" to select which have no facet information to display.

TODO: Can the paginator (or solr query) be extended to account for the number of unique facet values instead of `numDocs`?

straleyb:
Fixes #1349 
Fixes #1153 

This limits the max pages to the amount of records a user can see. It prunes out any records the user isnt allowed to view and stops the max pages at that value.

### DIFF
--- a/app/search_builders/scholars_archive/facet_modal_search_builder.rb
+++ b/app/search_builders/scholars_archive/facet_modal_search_builder.rb
@@ -1,17 +1,29 @@
 module ScholarsArchive
   class FacetModalSearchBuilder
+    #def admin_records(facet_string)
+    #  solr_service.get(facet_search_string(facet_string), :rows => 1000000)["response"]["docs"]
+    #end
+    #
+    #def group_records(facet_string, group, visibility: nil)
+    #  solr_service.get(facet_search_string(facet_string) + " AND " +
+    #                   read_access_group_string(group) + " AND " +
+    #                   visibility_string(visibility), :rows => 1000000)["response"]["docs"]
+    #end
+    #
+    #def edit_access_records(facet_string, username)
+    #  solr_service.get(facet_search_string(facet_string) + " AND " + edit_access_string(username), :rows => 1000000)["response"]["docs"]
+    #end
+
     def admin_records(facet_string)
-      solr_service.get(facet_search_string(facet_string), :rows => 1000000)["response"]["docs"]
+      "(" + [facet_search_string(facet_string), not_filesets].join(" AND ") + ")"
     end
 
     def group_records(facet_string, group, visibility: nil)
-      solr_service.get(facet_search_string(facet_string) + " AND " +
-                       read_access_group_string(group) + " AND " +
-                       visibility_string(visibility), :rows => 1000000)["response"]["docs"]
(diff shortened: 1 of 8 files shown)


app/search_builders/scholars_archive/facet_modal_search_builder.rb
```diff
@@ -0,0 +1,41 @@
+module ScholarsArchive
+  class FacetModalSearchBuilder
+    def admin_records(facet_string)
+      "(" + [facet_search_string(facet_string), not_filesets].join(" AND ") + ")"
```
revgum: Nitpicky, for sure.. but from a readability perspective could you use string interpolation for these methods instead of concatenation?

```
clauses = [facet_search_string(facet_string), not_filesets].join(" AND ")
"(#{clauses})"
```

app/services/scholars_archive/records_by_user_group_and_visibility.rb
```diff
@@ -0,0 +1,43 @@
...
+          query_strings << public_search(facet, group)
+        end
+      else
+        current_user.groups.each do |group|
+          query_strings << public_search(facet, group)
+          query_strings << authenticated_search(facet, group)
+        end
+      end
+      query_strings << other_owned_records(facet, current_user.username)
+      ActiveFedora::SolrService.get( query_strings.join(" OR "), :rows => 1000000 )["response"]["docs"].map { |item| item[facet.key.gsub("sim", "tesim")] }.uniq
```
revgum: Move this into the search_builder to keep that functionality together

straleyb: Id rather keep this out of the search builder. In hyrax, the search builders generate the query strings, but they query elsewhere. The query happens usually in a service of some kind.

straleyb: https://github.com/samvera/hyrax/blob/21964e1de237c560136b0136738612b258587552/app/services/hyrax/work_query_service.rb

revgum: https://github.com/samvera/hyrax/blob/21964e1de237c560136b0136738612b258587552/app/services/hyrax/work_query_service.rb#L51

revgum: So as I follow it, then you should remove `solr_service` from the builder...?

straleyb: Yeah, looks like thats some left over code from a previous iteration. Ill update.

PR #1530, opened 2018-07-26
Makes all fields available to admin

### LINKED ISSUE #1504: Metadata fields should always be editable for Administrators
Currently, some metadata fields are not editable (when values exist, they're locked as readonly):
- [ ] Location
- [ ] Related Items

Example:
https://ir.library.oregonstate.edu/concern/technical_reports/5q47rt93r/edit?locale=en

straleyb:
Fixes #1504 

The fields are unordered at this moment. If we want order, this will take a bit more work.

### DIFF
--- a/app/forms/scholars_archive/article_work_form_behavior.rb
+++ b/app/forms/scholars_archive/article_work_form_behavior.rb
@@ -5,12 +5,20 @@ module ScholarsArchive
       self.terms += [:resource_type, :editor, :has_volume, :has_number, :conference_location, :conference_name, :conference_section, :has_journal, :is_referenced_by, :web_of_science_uid]
 
       def primary_terms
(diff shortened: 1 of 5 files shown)


vantuyls:
@straleyb when did we have a conversation about ordering for admin not being important?

wickr:
This PR introduces several new problems:

* It's not just ordering, it's the primary/secondary term splits are all massively different, and any grouping (like dates) is all gone. Also gone is any connection to what a non-admin user sees and I can see this making troubleshooting or assistance more difficult.

* Custom Form views for Dates, Nested Geo, etc. are broken: 
sa-form-dates
sa-form-nestedgeo

* Some fields appear twice on same form. On Article form: Web of Science Uid, Identifier, Description, Dspace community, Dspace collection, Funding Body, Source, Keyword, Conference Name, Conference Section/Track, Conference Location, are all on there twice.

PR #1534, opened 2018-07-27
create template for SA issues

vantuyls:
template for issues, offering guidance for what types of information should be included.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,10 @@
+
+## Brief description of the issue
+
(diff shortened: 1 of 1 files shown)


wickr:
How about expanding that some more to something like the Hyrax template? https://github.com/samvera/hyrax/blob/master/.github/ISSUE_TEMPLATE.md

I like the Expected/Actual Behavior breakdown and Steps to Reproduce. Related is good too. We may not need a Rationale section.

vantuyls:
this is  dumb

Issue #1500, opened 2018-07-17
Implement ordering on Creator metadata field


vantuyls:
Approach is to make Creator a nested resource with type, order, and value as nested elements. Things to consider:
- [x] order is an integer
- [x] 0 index
- [x] value is a string
- [x] order validation: no duplicate index number
- [x] reindex order on save/submit
- [x] UI to reorder creators - similar to what we see on homepage for featured researcher, but can well-handle 35 or more creators [example](https://ir.library.oregonstate.edu/concern/articles/8910jw33p)
- [x] javascript validation in the UI (e.g. to prevent duplicates)?

vantuyls:
related sub-issues:
https://github.com/osulp/Scholars-Archive/issues/1542
https://github.com/osulp/Scholars-Archive/issues/1543
https://github.com/osulp/Scholars-Archive/issues/1544
https://github.com/osulp/Scholars-Archive/issues/1545

luisgreg99:
fixed by https://github.com/osulp/Scholars-Archive/pull/1554

PR #1554, opened 2018-07-31
Ordered Creators

straleyb:
fixes issue #1500 

This is the beginnings of the input to generate the fields in creator that will be swappable. It uses the featured works html and a new custom input to do this.


app/assets/javascripts/nested_field.js.coffee
```diff
@@ -54,11 +54,59 @@
...
+      # remove id
+      item.find("input[id$='_id']").remove()
+
+      # get child id
+      id = getLastChildId(item)
+
+      if id != -1
+        html = item.html()
+        time = new Date().getTime() # use current time to make sure the new id is unique
+        html = html.replace(new RegExp("\\["+id+"\\]", 'g'),'['+(id+time)+']')
```
revgum: @straleyb I believe these RegEx still need to be replaced, do you recall our work above on lines 60ish?

straleyb: Yeah, about changing the props instead of replacing the whole html. I can translate that work down into this if block as well.

luisgreg99:
Demo 1: 
nested-ordered-creator-1

PR #1576, opened 2018-08-10
Feature/up and down arrow

### LINKED ISSUE #1542: Create UI for Reordering Creator on Form
### Summary

Need to implement the UI for the ingest/edit forms for reordering creator. The UI used, currently, to reorder the Featured Works on the homepage. Specifics:
* UI should be no bigger than what we see on the Creator field currently
* Include up and down arrows
* Reindexing based on what is on the form after a change - drop, button click, and on form submit
* Use an ordered list element for creator

### Related work

#1500

straleyb:
fixes #1542

### DIFF
--- a/app/assets/javascripts/ordered_creator.js
+++ b/app/assets/javascripts/ordered_creator.js
@@ -5,6 +5,16 @@ function orderedDragAndDrop(selector) {
   });
 }
 
+function bindUpDownArrow(mutationsList) {
+  $('.ordered-up-arrow').on('click', function(e) {
+    $(this).parent().insertBefore($(this).parent().prev());
+  });
+  $('.ordered-down-arrow').on('click', function(e) {
+    $(this).parent().insertAfter($(this).parent().next());
+  });
+  reindexNestedOrderedField();
+}
+
 function reindexNestedOrderedField(mutationsList) {
   hidden_index_selectors = $(".ordered-field-container .index");
   hidden_index_selectors.each(function (index) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/ordered_creator.js
```diff
@@ -1,12 +1,45 @@
...
-      hidden_index_selectors[index].value = index;
-    });
+  selector.nestable({ maxDepth: 0, rootClass: 'ordered-field-container' });
+  selector.on('change', function (event) {
+    reindexNestedOrderedField();
+  });
+}
+
+function reindexNestedOrderedField(mutationsList) {
+  hidden_index_selectors = $($(".ordered-field-container .index"));
```
revgum: syntax issue is still here

app/assets/javascripts/ordered_creator.js
```diff
@@ -1,12 +1,45 @@
...
+  hidden_index_selectors.each(function (index) {
+    hidden_index_selectors[index].value = index;
   });
+  bindUpDownArrow();
```
revgum: I think this should be moved out of this method, then a new mutationobserver created to handle running this method, and then the method ran one time inside the Blacklight.onload (to bind all of the pre-existing rows)

app/assets/javascripts/ordered_creator.js
straleyb: This should be fixed with the merge of my other PR which was updated. Ill rebase this one on that after.

app/assets/javascripts/ordered_creator.js
straleyb: I was thinking of doing that and had it in there at one point. I wasnt sure if we wanted to keep creating mutation observers or just piggyback off the one.

PR #1579, opened 2018-08-13
Sorts on the show page

straleyb:
Reconfigures the nested_ordered_creator_label to include the index of the object, and adds a parser/sorter to parse index and sort labels based on index.

### DIFF
--- a/app/services/scholars_archive/ordered_parser_service.rb
+++ b/app/services/scholars_archive/ordered_parser_service.rb
@@ -2,11 +2,7 @@ module ScholarsArchive
   class OrderedParserService
     def self.parse(labels)
       labels ||= []
-      parsed_labels = []
-      puts labels
       sorted_labels = labels.map{ |label| build_array(label)}.sort_by{ |array| array[1] }.map{ |array| array[0] }
-      puts sorted_labels
-      sorted_labels
     end
 
     private
--- a/app/services/scholars_archive/ordered_parser_service.rb
+++ b/app/services/scholars_archive/ordered_parser_service.rb
@@ -1,0 +1,15 @@
+module ScholarsArchive
+  class OrderedParserService
+    def self.parse(labels)
+      labels ||= []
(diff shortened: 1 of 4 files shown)


app/services/scholars_archive/ordered_parser_service.rb
```diff
@@ -0,0 +1,15 @@
...
+  class OrderedParserService
+    def self.parse(labels)
+      labels ||= []
+      parsed_labels = []
+      sorted_labels = labels.map{ |label| build_array(label)}.sort_by{ |array| array[1] }.map{ |array| array[0] }
+    end
+
+    private
+
+    def build_array(label)
```
luisgreg99: It looks like this should be 
```
def self.build_array(label)
```
to prevent `undefined method build_array' for ScholarsArchive::OrderedParserService:Class`

straleyb: Yeah, I needed `self.` :P

app/services/scholars_archive/ordered_parser_service.rb
```diff
@@ -0,0 +1,18 @@
+module ScholarsArchive
+  class OrderedParserService
+    def self.parse(labels)
+      labels ||= []
+      parsed_labels = []
```
luisgreg99: It looks like this variable is not being used anywhere, so I think we can remove it.

app/services/scholars_archive/ordered_parser_service.rb
```diff
@@ -0,0 +1,18 @@
...
+      puts labels
+      sorted_labels = labels.map{ |label| build_array(label)}.sort_by{ |array| array[1] }.map{ |array| array[0] }
+      puts sorted_labels
```
luisgreg99: I think we can remove the lines `puts labels` and `puts sorted_labels` too. Did you put them there to inspect the values only? @straleyb

PR #1589, opened 2018-08-15
Rake task for migrating. Beginning work

straleyb:
This is the beginnings of the rake task. Need to set up reporting but wanted to get this in front of people.


lib/tasks/migrate_creators.rake
```diff
@@ -0,0 +1,68 @@
...
+          logger.warning("MISSMATCH_TYPE: MISSING CREATOR\n")
+          logger.warning("Name from DSpace was not found on this work.\n")
+          logger.warning("WORK_ID: #{work.id}\n")
+          logger.warning("CREATORS: #{work.creators}\n")
+          logger.warning("DSPACE_CREATOR: #{row["text_value"]}\n")
+          logger.warning("======================================================================================================")
+      end
+
+      # Add the current name and place as a nested ordered creator to the work.
+      work.nested_ordered_creators << NestedOrderedCreator.create(creator:row["text_value"], index: row["place"] - 1)
```
luisgreg99: I don't think `create` is defined for `NestedOrderedCreator`. To add new creator items, I usually do something like:
```
work.nested_ordered_creator_attributes = [
  {
    :index => "0",
    :creator => "Creator 0"
  },
  {
    :index => "1",
    :creator => "Creator 1"
  }
]
work.save
```
There are more examples in the model spec for `Default` type at https://github.com/osulp/Scholars-Archive/blob/master/spec/models/default_spec.rb#L232

lib/tasks/migrate_creators.rake
```diff
@@ -0,0 +1,68 @@
...
+    docs.each do |doc|
+
+      # Find work based on ID
+      work = ActiveFedora::Base.find(doc["id"]) 
+
+      # Iterate over the creators
+      work.creators.each_with_index do |creator, i|
+
+        # Translate creators over to nested ordered creators
+        work.nested_ordered_creators << NestedOrderedCreator.create(creator: creator, index: i)
```
luisgreg99: Same here regarding `create`. Also, I think it would be a good idea to rescue errors if something goes wrong and send them to the logs for reference. To rescue errors, I usually do something like:
```
begin
  work = work_model.find(doc["id"])

  # add or update attributes, etc

  if work.save
    logger.info "\t\t update for work id #{doc["id"]} completed successfully"
  else
    logger.info "\t\t failed to update work id #{doc["id"]} on save"
  end

rescue => e
  logger.info "\t\t failed to update work id #{doc["id"]}, error found:"
  logger.info "\t\t #{e.message}"
end
```

PR #1613, opened 2018-08-23
Adds ordering to related items and title. Separate from the refactor.

straleyb:


app/assets/javascripts/nested_field.js.coffee
```diff
@@ -82,20 +82,38 @@
       $(e).prop('id', id)
       return
 
+    allEmptyItems = (items) ->
```
revgum: This sort of method could be simplified with something along the lines of 

```
text_items = items.find('input:text')
empty_items = text_items.filter (e) -> $(e).val().length === 0
return empty_items.length === text_items.length

File: app/inputs/multi_value_ordered_input.rb [hunk 2]
```diff
@@ -66,7 +77,11 @@ def build_field(value, index)
       destroy_input = @builder.text_field(:_destroy, destroy_options)
     end
 
-    nested_item = "#{input_creator}#{input_index}"
+    nested_item = "#{input_creator}#{input_creator_2}#{input_index}"
+
+    puts "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
```
revgum: Don't forget t remove the puts or replace with Rails.logger

app/inputs/multi_value_ordered_input.rb
```diff
@@ -105,6 +120,33 @@ def build_creator_options(value, index)
     options
   end
 
+  def build_title_options(value, index)
+    creator_value = value.title.first
```
revgum: `creator_value` variable names don't make sense for these new methods

app/inputs/multi_value_ordered_input.rb
```diff
@@ -128,6 +170,7 @@ def nested_field_id(property, index)
 
   def collection
     @collection ||= begin
+      puts object[attribute_name].to_s
```
revgum: don't forget to remove this or replace with Rails.logger

PR #1622, opened 2018-08-27
updated models to use nested ordered title for works

### LINKED ISSUE #1618: Title may have a system level requirement and removing it from the form causes invalid form
### Descriptive summary

When you remove the core metadata title, it doesn't allow the form to be submitted. Some more digging needs to happen about how to stop the form from requiring and validating title and use ordered title instead.

luisgreg99:
fixes #1618


app/models/default.rb
```diff
@@ -12,12 +12,21 @@ class Default < ActiveFedora::Base
...
 
   self.validates_with ScholarsArchive::Validators::OtherOptionDegreeValidator
   self.validates_with ScholarsArchive::Validators::OtherAffiliationValidator
   self.validates_with ScholarsArchive::Validators::NestedRelatedItemsValidator
 
+  def to_s
+    if title.present?
+      title
+    elsif nested_ordered_title.first.title.present?
+      nested_ordered_title.select { |i| i.title.present? }.map{|i| (i.instance_of? NestedOrderedTitle) ? [i.title.first, i.index.first] : i }.select(&:present?).sort_by{ |array| array[1] }.map{ |array| array[0] }.first
```
revgum: Shouldn't all values in nested_ordered_title be instances of NestedOrderedTitle class?

Could these selects, maps, sort be moved to the NestedOrderedTitle class to some sort of clear method like `ordered_titles` which you can pull the `.first` of?

luisgreg99: @revgum All values in nested_ordered_title should be instances of NestedOrderedTitle on `edit`, but it might not be on the `new` forms. I believe the issue on values on new forms not being consistent has recently been fixed by https://github.com/osulp/Scholars-Archive/pull/1620 I think it would be a good idea to double check in this step. 

I've moved the title logic to a model concern `HasNestedOrderedProperties`, including the new method `ordered_titles` to simplify the steps.

PR #1627, opened 2018-08-28
Feature/add all other ordered fields

straleyb:


app/forms/scholars_archive/nested_behavior.rb
```diff
@@ -7,13 +7,20 @@ module NestedBehavior
       delegate :nested_geo_attributes=, :to => :model
       delegate :nested_related_items_attributes=, :to => :model
       delegate :nested_ordered_creator_attributes=, :to => :model
+      delegate :nested_ordered_contirbutor_attributes=, :to => :model
```
revgum: spelling error

app/inputs/multi_value_ordered_input.rb
```diff
@@ -52,14 +52,23 @@ def build_field(value, index)
     end
 
     if value.is_a?(NestedOrderedCreator)
-      creator_options = build_creator_options(value, index)
+      creator_options = build_item_options(value, index, :creator)
       input_creator = @builder.text_field(:creator, creator_options)
     elsif value.is_a?(NestedOrderedTitle)
-      creator_options = build_title_options(value, index)
+      creator_options = build_item_options(value, index, :title)
```
revgum: creator_options as a variable name doesn't make sense here and in following lines

app/assets/javascripts/nested_field.js.coffee
```diff
@@ -130,7 +130,17 @@
 
       $(field_selector).append($(reindex_ordered_list))
 
-    resetNestedFieldItems('.nested-ordered-creator', false)
-    resetNestedFieldItems('.nested-ordered-title', false)
-    resetNestedFieldItems('.nested-ordered-related-items', true)
+    nested_fields = ['.nested-ordered-creator', 
+                     'nested-ordered-title',
```
revgum: selectors missing period

app/assets/javascripts/ordered_creator.js
```diff
@@ -27,25 +27,34 @@ function swapDown(selector) {
...
-  hidden_index_selectors.each(function (index) {
-    hidden_index_selectors[index].value = index;
-  });
 
-  hidden_index_selectors = $(".nested-ordered-related-items .index");
-  hidden_index_selectors.each(function (index) {
-    hidden_index_selectors[index].value = index;
-  });
+  nested_fields = ['.nested-ordered-creator', 
+                     'nested-ordered-title',
```
revgum: selectors missing period

app/assets/javascripts/ordered_creator.js
```diff
@@ -27,25 +27,34 @@ function swapDown(selector) {
...
+  })
 }
 
 Blacklight.onLoad(function () {
-  orderedDragAndDrop($('.nested-ordered-related-items'));
-  orderedDragAndDrop($('.nested-ordered-title'));
-  orderedDragAndDrop($('.nested-ordered-creator'));
+
+  nested_fields = ['.nested-ordered-creator', 
+                     'nested-ordered-title',
```
revgum: selectors missing period

straleyb:
Still need to swap over labels so the form has the right titles for the fields, check the js to make sure its firing and working proper, and do spot checking to make sure the new fields are fully functional. Had a first round off feedback and made fixes.

app/inputs/multi_value_ordered_input.rb
```diff
@@ -59,6 +59,31 @@ def build_field(value, index)
       index = value.object_id
     end
 
+    if value.is_a?(NestedOrderedCreator)
+      item_options = build_item_options(value, index, :creator)
+      input_creator = @builder.text_field(:creator, item_options)
+    elsif value.is_a?(NestedOrderedTitle)
+      item_options = build_item_options(value, index, :title)
+      input_creator = @builder.text_field(:title, item_options)
```
luisgreg99: I think it would be a good idea to rename these variables to match the nested field names for consistency, so the line above coule be: 
```
input_title = @builder.text_field(:title, item_options)
```

I did some of of these changes in a previous PR, see https://github.com/osulp/Scholars-Archive/pull/1628/files#diff-057c8e520a5ab3d69344bf096f4e5553R69

I basically moved the logic to a new method `build_nested_item(value,index)` and renamed the variables there. This method returns the appropriate inputs based on the nested models.

app/models/nested_ordered_abstract.rb
```diff
@@ -0,0 +1,31 @@
+class NestedOrderedAbstract < ActiveTriples::Resource
+  # Usage notes and expectations can be found in the Metadata Application Profile:
+  #   https://docs.google.com/spreadsheets/d/1koKjV7bjn7v4r5a3gsowEimljHiAwbwuOgjHe7FEtuw/edit?usp=sharing
+
+  property :index, predicate: ::RDF::Vocab::DC.identifier
+  property :abstract, predicate: ::RDF::Vocab::DC.creator
```
luisgreg99: I believe the predicate for `:abstract` should be `::RDF::Vocab::DC.abstract`.

app/models/nested_ordered_contributor.rb
```diff
@@ -0,0 +1,31 @@
+class NestedOrderedContributor < ActiveTriples::Resource
+  # Usage notes and expectations can be found in the Metadata Application Profile:
+  #   https://docs.google.com/spreadsheets/d/1koKjV7bjn7v4r5a3gsowEimljHiAwbwuOgjHe7FEtuw/edit?usp=sharing
+
+  property :index, predicate: ::RDF::Vocab::DC.identifier
+  property :contributor, predicate: ::RDF::Vocab::DC.creator
```
luisgreg99: For `:contributor`, I think it's `::RDF::Vocab::DC11.contributor`.

app/models/nested_ordered_description.rb
```diff
@@ -0,0 +1,31 @@
+class NestedOrderedDescription < ActiveTriples::Resource
+  # Usage notes and expectations can be found in the Metadata Application Profile:
+  #   https://docs.google.com/spreadsheets/d/1koKjV7bjn7v4r5a3gsowEimljHiAwbwuOgjHe7FEtuw/edit?usp=sharing
+
+  property :index, predicate: ::RDF::Vocab::DC.identifier
+  property :description, predicate: ::RDF::Vocab::DC.creator
```
luisgreg99: For `:description`, I think it's `::RDF::Vocab::DC11.description`.

app/models/solr_document.rb
```diff
@@ -91,6 +91,22 @@ def title
     self[Solrizer.solr_name('title', :stored_searchable)] || nested_ordered_title_label || []
   end
 
+  def description
```
luisgreg99: We could extend this method to return existing unordered values if available as a fallback when the ordered values are not available yet. I think this would help during migration so that we can still display those values.

For example, we can do something like:
```
def description
     nested_ordered_description_label.present? ? nested_ordered_description_label : self[Solrizer.solr_name('description', :stored_searchable)]
  end
```

app/assets/javascripts/ordered_creator.js
```diff
@@ -27,25 +27,34 @@ function swapDown(selector) {
...
-  hidden_index_selectors = $(".nested-ordered-related-items .index");
-  hidden_index_selectors.each(function (index) {
-    hidden_index_selectors[index].value = index;
-  });
+  nested_fields = ['.nested-ordered-creator .index', 
+                     '.nested-ordered-title .index', 
+                     '.nested-ordered-abstract .index', 
+                     '.nested-ordered-contributor .index', 
+                     '.nested-ordered-description .index', 
+                     '.nested-ordered-related-item .indexs']
```
revgum: is the trailing s `indexs` a bug? I think the selector is supposed to be `.nested-ordered-related-items .index`?

app/forms/scholars_archive/default_work_form_behavior.rb
```diff
@@ -14,19 +14,19 @@ module DefaultWorkFormBehavior
...
       end
 
       def secondary_terms
-        t = [:nested_related_items, :hydrologic_unit_code, :geo_section, :funding_statement, :publisher, :peerreviewed, :conference_name, :conference_section, :conference_location, :language, :file_format, :file_extent, :digitization_spec, :replaces, :additional_information, :isbn, :issn]
-        t << [:keyword, :source, :funding_body, :dspace_community, :dspace_collection, :description, :identifier] if current_ability.current_user.admin?
+        t = [:nested_related_items, :hydrologic_unit_code, :geo_section, :funding_statement, :publisher, :peerreviewed, :conference_name, :conference_section, :conference_location, :language, :file_format, :file_extent, :digitization_spec, :replaces, :nested_ordered_additional_information, :isbn, :issn]
+        t << [:keyword, :source, :funding_body, :dspace_community, :dspace_collection, :nested_ordered_additional_information, :identifier] if current_ability.current_user.admin?
```
luisgreg99: It looks like this line removes `:description` from the form. I remember @simholt mentioned that we should still keep it in the form for admin users only. I think we can leave it as:
```
t << [:keyword, :source, :funding_body, :dspace_community, :dspace_collection, :description, :identifier] if current_ability.current_user.admin?
```

PR #1643, opened 2018-09-05
Feature/remove label help text

### LINKED ISSUE #1641: Remove placeholder text for all but related items
### Descriptive summary

Its completely unnecessary to have the placeholder text on any of the fields besides the related items. We need to remove those.

straleyb:
Fixes #1641


app/inputs/multi_value_ordered_input.rb
```diff
@@ -142,12 +142,12 @@ def build_destroy_options(value, index)
...
   end
 
-  def build_item_options(value, index, method)
+  def build_item_options(value, index, method, placeholder)
     creator_value = value.send(method).first
     options = build_field_options(creator_value, index)
     options[:name] = nested_field_name(method.to_s, index)
     options[:id] = nested_field_id(method.to_s, index)
-    options[:placeholder] = 'Label'
+    options[:placeholder] = placeholder
```
luisgreg99: @straleyb what if we do something like
```
options[:placeholder] = method.to_s.titleize if prop_with_labels.include?(method)
```
And then define the list of properties that need labels:
```
def prop_with_labels
  [:label, :related_url]
end
```
I’m thinking in this way we can keep the build_item_options method simpler.

PR #1664, opened 2018-09-10
Feature/text area style

straleyb:


straleyb:
screen shot 2018-09-10 at 1 17 10 pm

straleyb:
When you extend the text area, the buttons and handle stay put. Dont increase in size or anything. Are simply static.

PR #1675, opened 2018-09-12
fixes search

### LINKED ISSUE #1672: Unlinkify additional information and abstract
### Descriptive summary

Both abstract and additional information dont need to be links to searches within SA.

### LINKED ISSUE #1673: Quick search for nested_ordered_items that are public arent searching
### Descriptive summary

On the show page of a public work. If you have a work that has a nested ordered item, and you click the quick search link, it isnt searching properly.

straleyb:
Fixes #1673 Fixes #1672

### DIFF
--- a/app/views/hyrax/base/_attribute_rows.html.erb
+++ b/app/views/hyrax/base/_attribute_rows.html.erb
@@ -1,9 +1,9 @@
-<%= presenter.attribute_to_html(:nested_ordered_title_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_title_label', itemprop_option: 'url', label: 'Title') %>
+<%= presenter.attribute_to_html(:nested_ordered_title_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_title_label_ssim', itemprop_option: 'url', label: 'Title') %>
 <%= presenter.attribute_to_html(:alt_title, render_as: :external_link, label: t('simple_form.labels.defaults.alt_title')) %>
-<%= presenter.attribute_to_html(:nested_ordered_creator_label, render_as: :scholars_archive_nested, search_field: 'nested_creator', itemprop_option: 'url', label: "Creator") %>
+<%= presenter.attribute_to_html(:nested_ordered_creator_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_creator_label_ssim', itemprop_option: 'url', label: "Creator") %>
 <%#= presenter.attribute_to_html(:creator, render_as: :search_and_external_link) %>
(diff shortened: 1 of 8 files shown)


app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,9 +1,9 @@
-<%= presenter.attribute_to_html(:nested_ordered_title_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_title_label', itemprop_option: 'title', label: 'Title') %>
+<%= presenter.attribute_to_html(:nested_ordered_title_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_title_label', itemprop_option: 'url', label: 'Title') %>
 <%= presenter.attribute_to_html(:alt_title, render_as: :external_link, label: t('simple_form.labels.defaults.alt_title')) %>
-<%= presenter.attribute_to_html(:nested_ordered_creator_label, render_as: :scholars_archive_nested, search_field: 'nested_ordered_creator_label', itemprop_option: 'creator', label: "Creator") %>
+<%= presenter.attribute_to_html(:nested_ordered_creator_label, render_as: :scholars_archive_nested, search_field: 'nested_creator', itemprop_option: 'url', label: "Creator") %>
```
luisgreg99: Does this change maintain the order set in `nested_ordered_creator`? I think it would be a good idea to define a spec for this case in https://github.com/osulp/Scholars-Archive/blob/master/spec/renderers/hyrax/renderers/scholars_archive_nested_attribute_renderer_spec.rb to check the order as well in the renderer.

straleyb:
working-quicklink

luisgreg99:
@straleyb Nice, I just have one observation. It looks like the label in the search tag container next to "Filtering By" shows the solr name  for the field being searched. See screenshot below:

image

I think it should say something like `Contributor > contribute` instead of `Nested ordered contributor label ssim > contribute`

The label could be set in the config section in the catalog controller. For example, I remember for related items, I believe it's set in:
https://github.com/osulp/Scholars-Archive/blob/master/app/controllers/catalog_controller.rb#L236

straleyb:
Odd because I do have those set https://github.com/osulp/Scholars-Archive/blob/master/app/controllers/catalog_controller.rb#L261
and
https://github.com/osulp/Scholars-Archive/blob/master/app/controllers/catalog_controller.rb#L269

straleyb:
Maybe its not the right label? Ill dive into it and check it out.

straleyb:
see https://github.com/osulp/Scholars-Archive/issues/1679

PR #1758, opened 2018-11-07
Fixes collection searching

### LINKED ISSUE #1739: Cannot add items to collection manually/individually
### Expected behavior
Using the button on an item's show page OR the Relationships tab in an item's edit page, should be able to add an item to a collection by typing the name of the collection in the search box. This worked two weeks ago.

### Actual behavior
No matter what I type, no collections are returned in the search.

screen shot 2018-10-30 at 2 54 37 pm

### Steps to reproduce the behavior
Choose any item and attempt to add it to a collection, such as Pacific Northwest Insect Management Conference.

straleyb:
Fixes #1739 

The issue was that it was doing a solr search for the title, which on collections is still indexed under title_tesim, but we removed it from the all fields query so it wasnt searching for the proper title. Ill provide a screenshot and gif of it, then after that it should be good to go.


straleyb:
screen shot 2018-11-07 at 7 59 26 am

straleyb:
collection search

simholt:
Yaaaaaaaay!

wickr:
One concern with this is we still have work titles in `title_tesim` so we'll have them in the "all fields" search field twice with putting the field back in. Probably the best option going forward is to clear those out.

Also, FileSet titles (usually filename) are also stored in `title_tesim` so this would bring those back as searchable.

wickr:
Example Solr result:
```
"title_tesim":["Restoring forest health : fire and thinning effects on mixed-conifer forests"],

"nested_ordered_title_tesim":["http://127.0.0.1:8080/fcrepo/rest/prod/ks/65/hd/98/ks65hd98x#nested_ordered_titleg70083383978800"],

"nested_ordered_title_label_ssim":["Restoring forest health : fire and thinning effects on mixed-conifer forests$0"],
"nested_ordered_title_label_tesim":["Restoring forest health : fire and thinning effects on mixed-conifer forests$0"],
```

straleyb:
Right, but in the solr document, when its grabbing title, it always looks at the nested ordered title first
```
def title
     nested_ordered_title_label.present? ? nested_ordered_title_label : self[Solrizer.solr_name('title', :stored_searchable)] || []
  end
```
so we use nested ordered title if it has it, and if it doesnt then we use title instead.

So it shouldnt ever pull from the wrong field.

wickr:
Yes but for search purposes, the title is coming in to "all fields" from 2 separate search indexes and could unnecessarily boost titles. If we don't need `title_tesim` values for works anymore we should get rid of them, also to avoid drift as some titles get updated but both fields don't update in sync.

straleyb:
So then is the question, do collections need ordered titles? if so, then this is a bigger task.

wickr:
Just to record what we discussed in standup, no need for nested ordered titles on Collections.

PR #1761, opened 2018-11-07
redirects a work to its proper route

### LINKED ISSUE #1696: Changing work from Default model to Article model causes error at old URL
### Descriptive summary

One of the worktypes changes needed for some works is changing from the Default model to the Article model. This has been completed for over 3400 works, and seemed to be ok. However, the 'old' Default model access URL:
https://ir.library.oregonstate.edu/concern/defaults/qv33rx600

throws an error: ```ActionView::Template::Error (undefined method `has_journal' for #<DefaultPresenter:0x00007f0200337648>):```

because the Default presenter doesn't know about `has_journal` now that the work is an Article. 

So anyone with the old link (which we used to say was the Citeable URL) or coming from search engines will now get the 'Ack SA is temporarily broken' error instead of any type of work show page.

This is another downside of having the work's model/type in the URL, and Hyrax allowing any valid presenter to try and display a work. ( https://ir.library.oregonstate.edu/concern/purchased_e_resources/qv33rx600 works ok because it includes article metadata and presenter, while https://ir.library.oregonstate.edu/concern/undergraduate_thesis_or_projects/qv33rx600  does not and also errors). 

Possible solutions are adding the Article-specific properties to the Default model so the error goes away and the work is viewable, or putting in another redirect system that will redirect the old, Default model URL to the new Article URL for any work changed. (That solution could also potentially work for #1378)

### Expected behavior
The Default model URL should still present the work and not error: https://ir.library.oregonstate.edu/concern/defaults/qv33rx600

### Actual behavior
When visiting the Default model URL after worktype change, get error "Ack! Looks like ScholarsArchive@OSU is temporarily broken."

revgum:
fixes #1696 

See slack channel for video demonstrating it in action

### DIFF
--- a/app/controllers/concerns/scholars_archive/works_controller_behavior.rb
+++ b/app/controllers/concerns/scholars_archive/works_controller_behavior.rb
@@ -3,7 +3,7 @@ module ScholarsArchive
     extend ActiveSupport::Concern
     include Hyrax::WorksControllerBehavior
     included do
-      before_action :redirect_mismatched_work
+      before_action :redirect_mismatched_work, only: [:show]
       before_action :migrate_work_in_place, only: [:edit, :show]
 
       def redirect_mismatched_work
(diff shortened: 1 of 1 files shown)


wickr:
Looks good, only thing I would add is to specify a 301 HTTP status code for Moved Permanently, instead of the default 302 Found

PR #1762, opened 2018-11-07
displays representative media for datasets

### LINKED ISSUE #1713: Add representative media fields to dataset worktype
Items in the dataset worktype do not have any of the representative media fields (see below). This would be useful, for example to set the readme file as the default download.

screen shot 2018-10-16 at 2 39 12 pm

straleyb:
Fixes #1713

### DIFF
--- a/app/views/hyrax/base/_form_metadata.html.erb
+++ b/app/views/hyrax/base/_form_metadata.html.erb
@@ -3,6 +3,7 @@
             <%= render_edit_field_partial(term, f: f) %>
           <% end %>
         </div>
+        <%= render 'form_media', f: f %>
         <% if f.object.display_additional_fields? %>
           <%= link_to t('hyrax.works.form.additional_fields'),
                       '#extended-terms',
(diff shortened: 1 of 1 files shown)


straleyb:
screen shot 2018-11-07 at 1 31 57 pm

straleyb:
The issue here is that since datasets dont have a secondary term field, and only primary terms, it wouldn't display. We would have to hack some gross pieces of code to get this to show up, this changes less code and makes it so it shows up. I would suggest over-engineering it wouldn't be beneficial to us.

straleyb:
Because of this line of code
https://github.com/osulp/Scholars-Archive/pull/1762/files#diff-44a3c32ef53c83e70f3b16b5b982813dL6

revgum:
Add the modified `_form_metadata.html.erb` to `datasets` instead of `base` and revert the change to the partial in `base`

vantuyls:
i think this is where some discussion on the original issue would have been useful.

PR #1787, opened 2019-01-09
rubocop installed some fixes

straleyb:


Gemfile
```diff
@@ -92,6 +92,8 @@ gem 'riiif', '~> 1.1'
 # Security update
 gem 'rubyzip', '~> 1.2.2'
 
+gem 'rubocop'
```
revgum: These should only be in the development&test group

Issue #1796, opened 2019-01-23
Migrated creators 10+ out of order


simholt:
### Descriptive summary
After ordered fields were established, it turned out that order was not maintained for non-migrated items of 10+ terms, such as Creator. ([#1773](https://github.com/osulp/Scholars-Archive/issues/1773))

It is a problem for migrated items as well.

9 is fine
https://ir.library.oregonstate.edu/concern/articles/02870w72j
http://dspace-ir.library.oregonstate.edu/xmlui/handle/1957/36332?show=full

As before, 10 creators is the magic number -- out of order
https://ir.library.oregonstate.edu/concern/articles/bn999744w
http://dspace-ir.library.oregonstate.edu/xmlui/handle/1957/42664?show=full

13 creators -- out of order
https://ir.library.oregonstate.edu/concern/defaults/xg94hv61v
http://dspace-ir.library.oregonstate.edu/xmlui/handle/1957/36591?show=full

ALSO: Order was not maintained for Contributor, but there was only one item with 10+ entries so I fixed it manually.

### Related work

[#1773](https://github.com/osulp/Scholars-Archive/issues/1773)

simholt:
To make this more complicated, any name with diacritics in DSpace was brought into Hyrax twice -- once with unrecognized characters and again with proper diacritics.

![screen shot 2019-01-29 at 7 51 37 am](https://user-images.githubusercontent.com/33705514/51920672-baf40200-239a-11e9-8d98-d989c57392fe.png)
- - - 
This gives us two types of problems:

- Less than ten creators: the unrecognized character version is in the correct order and the diacritic one is kicked to the end of the list. Fixing this means deleting the wrong one. [Example](https://ir.library.oregonstate.edu/concern/articles/5138jg786)
![screen shot 2019-01-29 at 7 54 24 am](https://user-images.githubusercontent.com/33705514/51920887-2342e380-239b-11e9-8414-844629799689.png)

- Ten or more creators means the order is messed up in two ways: loss of order as listed in DSpace and duplication of names with diacritics. [Example](https://ir.library.oregonstate.edu/concern/articles/pv63g434h)

![screen shot 2019-01-29 at 8 00 35 am](https://user-images.githubusercontent.com/33705514/51921332-0a86fd80-239c-11e9-946b-5a2f886d6dcd.png)
- - - 
Other examples:
https://ir.library.oregonstate.edu/concern/articles/th83m1037
[...]

PR #1822, opened 2019-03-19
add task to get children from handles manually fixed

luisgreg99:
Related to https://github.com/osulp/Scholars-Archive/issues/1796 

Some works fixed manually have child works that still have unclean creators. My original tasks didn't take into account child works already fixed in https://github.com/osulp/Scholars-Archive/issues/1796#issuecomment-472152751 

This tasks gets a json file with handles, including child works that we can analyze 
```
"handle-url" => [{
  member_id: "ABCDE", 
  member_creators: {"hello","world"},
  member_model: "articles"
}]
```
Update:

I'm also adding another task R3 to generate `r3_handles.csv` of handles with child works that we could run as `scholars_archive:migrate_dspace_creator_member_cleanup`

### DIFF
--- a/lib/tasks/migrate_works_10_or_more_items.rake
+++ b/lib/tasks/migrate_works_10_or_more_items.rake
@@ -83,6 +83,46 @@ namespace :scholars_archive do
     end
   end
 
+  # Task 2 (a): Generate a Json with works that were already fixed manually, but
+  # contain child works
+  desc "Generate a Json with handles containing unclean creators already fixed manually and include children info"
+  task :get_json_of_handles_x_with_creators_and_children => :environment do
+    handles_x = File.join(Rails.root, 'tmp', 'creator_cleanup', 'handles_with_unclean_creators_fixed.json')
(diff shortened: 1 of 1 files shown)


lib/tasks/migrate_works_10_or_more_items.rake
```diff
@@ -230,6 +308,11 @@ namespace :scholars_archive do
     end
   end
 
+  def get_work_members(doc)
+    work = ActiveFedora::Base.find(doc.id) if doc.respond_to?('id')
+    work.members.reject { |m| m.class.to_s == 'FileSet' }.map {|c| {membed_id: c.id, member_creators: c.creator, member_model: c.has_model.first.underscore.pluralize}}
```
wickr: Looks like this should be `member`

luisgreg99: I've updated it to `member_id`

wickr: fixed

PR #1836, opened 2019-04-22
simple override to limit resource type choices

### LINKED ISSUE #1492: Restrict Resource Type for GTD deposits
GTD depositors should only see the following Resource Types:
- Dissertation
- Masters Thesis

### LINKED ISSUE #1496: Restrict Resource Type for HCT deposits
Honors College Work Type deposits should only offer the option of "Honors College Thesis" for Resource Type

straleyb:
Fixes #1496 
Fixes #1492 

Screen Shot 2019-04-22 at 2 49 47 PM
Screen Shot 2019-04-22 at 2 52 59 PM


spec/views/records/edit_fields/_resource_type.html.erb_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'rails_helper'
```
revgum: These aren't necessary since `.rspec` provides this by default.

straleyb: Im just trying to not break the way we have things now. all the other ones have it too. Im trying to balance going in and fixing everything and getting my work in

app/views/records/edit_fields/_resource_type.html.erb
```diff
@@ -0,0 +1,5 @@
+<% collection = Hyrax::ResourceTypesService.select_options %>
+<% collection.select! { |tuple| tuple.first.include? "Honors" } if f.object.model.class.to_s == "HonorsCollegeThesis" %>
+<% collection.select! { |opt| opt.first.include?("Dissertation") || opt.first.include?("Masters") } if f.object.model.class.to_s == "GraduateThesisOrDissertation" %>
```
luisgreg99: Would it be a good idea to move the select logic to a helper method?

straleyb: So @revgum and I discussed moving this piece of code into a helper and we decided against it. There were a few reasons why we decided against it, First being that if we move this code into a helper, when we still have to override this file and in retrospect would create more code. The second being I feel as though this code would end up being swept under the rug and forgotten in a helper. The last place I start looking for code is in a helper. Also this piece of code is used only once, here, so it wouldnt necessarily be fully utilized in the helper as well. It would just fill the helper file up more. 

We had discussed creating our own resource type select service and adding a method where we could ask for a single or a list of resource types depending on the model, but that was a less sustainable override than this one. Keeping all this functionality in one place on the view seemed in the end like the best option. 

We definitely thought about it but ultimately decided against it. Hope this helps clue you in on what we were thinking here.

PR #1842, opened 2019-04-24
First iteration of removing trailing and leading whitespace

straleyb:
So this is a a huge work in progess. Today I looked into a possible gem that was supposed to remove trailing and leading whitespaces from inputs, but it didnt work with active fedora object. So I needed to dig my heels in and figure out the best way to iterate over all the attributes, inspect the values and call #strip on all the string. 

This means ill have to skip over attributes like datetimes. It also means that ill have to dig into the objects such as Nested Related Items and such to see what their values are and call #strip on those if they are strings. 

There are two possibilities to do so, one is to hit the parameters as they come in and do all the work there, or do it on the attributes of the work before save and allow it to save naturally. This goes for the attributes. if this way proves to be more difficult then ill try the parameters as well.


app/services/scholars_archive/param_scrubber.rb
```diff
@@ -0,0 +1,55 @@
...
+          # set the stripped string on the params 
+          set_params(params, hash_key, attr, value)
+        end
+      end
+    end
+
+    private
+
+    def self.extract_hash_values(hash)
+      hash.each_pair do |key, value|
```
revgum: I'm surprised rubocop didn't warn on this, it seems like it would suggest

```
      hash.each_pair do |key, value|
        return strip_and_set(hash, key, value) if value.respond_to?(:strip)

        # recursively process nested hashes
        extract_hash_values(value)
      end
```

PR #1849, opened 2019-04-30
Feature/disable batch upload for non admin

### LINKED ISSUE #1472: Disable batch upload for non-admin users
batch is borked (skipping workflows), so we want to avoid people using it until Hyrax gets batch fixed.

straleyb:
Fixes #1472 

Screen Shot 2019-04-30 at 1 28 14 PM
Screen Shot 2019-04-30 at 1 27 10 PM

### DIFF
--- a/.rubocop_fix_me.yml
+++ b/.rubocop_fix_me.yml
@@ -988,6 +988,8 @@ RSpec/VerifiedDoubles:
     - 'spec/views/shared/_select_work_type_modal.html.erb_spec.rb'
     - 'spec/views/hyrax/base/_work_description.erb_spec.rb'
     - 'spec/views/hyrax/base/_form_instructions_spec.rb'
+    - 'spec/views/hyrax/base/_form.html.erb_spec.rb'
+    - 'spec/views/hyrax/my/works/index.html.erb_spec.rb'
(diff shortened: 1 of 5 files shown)


straleyb:
For sure, just wanted to get it in front of eyes before i did that. I hated having to override it. I wanted to see if I could use abilities and other things to drive it, but ultimately, I feel this is probably the best option. Just wanted to see if anyone else had anything to say about it.

wickr:
There's also a link to Batch on the edit form: "To create a separate work for each of the files, go to Batch upload" to suppress

.rubocop_fix_me.yml
```diff
@@ -988,6 +988,8 @@ RSpec/VerifiedDoubles:
     - 'spec/views/shared/_select_work_type_modal.html.erb_spec.rb'
     - 'spec/views/hyrax/base/_work_description.erb_spec.rb'
     - 'spec/views/hyrax/base/_form_instructions_spec.rb'
+    - 'spec/views/hyrax/base/_form.html.erb_spec.rb'
```
straleyb: I added these because trying to create some of the doubles for these objects using the proper "verified doubles" is a non trivial thing to do. This is how hyrax did their tests and im willing to stick with it for sanity sake.

PR #1856, opened 2019-05-06
reenable clamav

straleyb:


revgum:
You also need to address: https://github.com/osulp/Scholars-Archive/blob/master/config/initializers/clamav.rb

PR #1858, opened 2019-05-06
fixes day of embargo expiring issue

### LINKED ISSUE #1857: fix embargo task
So the problem is that when an embargo is expired on the same day as the embargo release date, the application doesn’t actually apply the visibility and embargo changes. Making matters worse, it unsets the release date, visibility before&after metadata on the embargo so the next time a embargo release is run it wouldn’t ever see this as having been expired either. One part of this logic problem is this evaluation considering that `Date.today < embargo_release_date`.. if it considered equality then this wouldn’t be as much of a problem (`Date.today <= embargo_release_date`). The flow works like this;

```
- work.embargo_visibility! : evaluates that the work is still `under_embargo?` and intends on maintaining the visibility settings
- `under_embargo?` evaluates to true because of the `Date.today < embargo_release_date` is false because Date.today EQUALS embargo_release_date
- then, `work.deactivate_embargo!` manually deactivates the embargo.. not setting the works visibility or anything.. just leaving them as-is
- then, the embargo is saved and the work is saved
```
Solutions:
If we're fine with the embargo releasing the day _after_ it expires, then:
just add `next if work.under_embargo?` after this line https://github.com/osulp/Scholars-Archive/blob/master/lib/scholars_archive/embargoes/embargo_releaser.rb#L12

Otherwise, then we monkeypatch the `active?` https://github.com/samvera/hydra-head/blob/master/hydra-access-controls/app/models/hydra/access_controls/embargo.rb#L14

straleyb:
Fixes #1857

### DIFF
--- a/lib/scholars_archive/embargoes/embargo_releaser.rb
+++ b/lib/scholars_archive/embargoes/embargo_releaser.rb
@@ -10,6 +10,7 @@ module ScholarsArchive::Embargoes
       expired_embargoes = Hyrax::EmbargoService.assets_with_expired_embargoes
       expired_embargoes.each do |embargo|
         work = ActiveFedora::Base.find(embargo.solr_document[:id])
+        next if work.under_embargo?
         logger.warn("Processing work id: #{work.id}")
         begin
           work.embargo_visibility!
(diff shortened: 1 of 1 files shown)


revgum:
@vantuyls Feedback on this to fix the rake task for embargo expiration or the other solution in #1857

PR #1867, opened 2019-05-14
[WIP] override indexer to change what is being indexed

straleyb:
This is an attempt to fix what is being indexed. The goal of this work is to get the metadata in line with which version of a file is the currently selected version. There are a couple of mechanisms that fit into place here. 

First is https://github.com/osulp/Scholars-Archive/compare/feature/DisplayProperThumbnail?expand=1#diff-27e6b51d035075f8e17abdcab956ae9cR1 . This was overridden to hook in our indexing service into it.
Second is https://github.com/osulp/Scholars-Archive/compare/feature/DisplayProperThumbnail?expand=1#diff-032d414b63bd91495cbd33302cba512eR3 which sets the file sets indexer to our indexer
Third is https://github.com/osulp/Scholars-Archive/compare/feature/DisplayProperThumbnail?expand=1#diff-d6262e8ed24efb309e60bfa592588a88R1

If we want to get the metadata in line with whats going on. We need to have methods that are able to figure out which version is chosen, then ask the solr document for the metadata from a specific version. This means we have to index the metadata per version that is ingested. There is still more work to be done here. 

This is the file that renders the data on the show page for a specific file set. 

https://github.com/samvera/hyrax/blob/726721ee5247e93feb2ac70ea4cac4d116dcedcb/app/views/hyrax/file_sets/_show_characterization_details.html.erb#L3

This is where it iterates over the fields and extracts them from the solr document and gets them into a chunk of data. 

https://github.com/samvera/hyrax/blob/09085ea50e84735c3aaea031759715b07334f260/app/presenters/hyrax/characterization_behavior.rb#L76

If we can get the solr document in line with the way we want the data to be shaped and have it be able to display the proper versions data in the characterization behavior, then we should be able to display version specific metadata on the file set. 

However, Currently im struggling to get our indexer hooked in. It knows that our indexer should be the one it uses, but `generate_solr_document` isnt being called on it, so it isnt indexing any of the data and not displaying the file set. 

If https://github.com/osulp/Scholars-Archive/compare/feature/DisplayProperThumbnail?expand=1#diff-3464031fc30a4eff06fb022cfd8a76d9R5 is replaced with the old code `include ::Hyrax::FileSetBehavior` it works, but with ours, it doesnt even call generate_solr_document. 

Once we can get that hooked in, we can get the proper data indexed by using the version info. Then when it retrieves the data, we can use version info to pull from the solr document and display the proper data.


straleyb:
Closing due to hyrax issue

PR #1884, opened 2019-06-06
Google scholar citation updates

luisgreg99:
Fixes https://github.com/osulp/Scholars-Archive/issues/1916

Google Scholar requires citation publication date meta tags to be in the format `Y/m/d` according to the indexing guidelines https://scholar.google.com/intl/en/scholar/inclusion.html#indexing.

- This update adds a `citation_publication_date` helper to format dates from the presenter to follow the guidelines. This one is used to format `date_uploaded`, which comes in the form `m/d/Y` and is a fallback when `date_created`, `date_issued`, and `date_accepted` are not available.
- It also skips empty or unsupported meta-tags like `citation_firstpage` and `citation_lastpage`.
- Robots.txt update: allow `catalog` to `Googlebot`

Notes/findings:

- [x] 1. Find out if Google Scholar accepts `Y/m`- Update: Yes
- [x] 2. Find out if Google Scholar accepts date intervals like `Y/m/d-Y/m/d` - Update: No
- [x] 3. If (2) is not possible, check with metadata group and see if we can cleanup publication date when given a date interval for `date_created`, `date_issued`, and `date_accepted` (these fields are used in `citation_publication_date` meta tag for Google Scholar).

### DIFF
--- a/app/services/scholars_archive/google_scholar_citation_service.rb
+++ b/app/services/scholars_archive/google_scholar_citation_service.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module ScholarsArchive
+  # Class to handle the expected format for citation publication dates as
+  # required in the indexing guidelines for configuring meta-tags:
+  # https://scholar.google.com/intl/en/scholar/inclusion.html#indexing
+  class GoogleScholarCitationService
+    attr_reader :input_date
+
+    def self.call(input_date)
+      new(input_date).call
+    end
+
+    def call
+      format_date
+    end
+
+    private
+
+    def initialize(input_date)
+      @edtf_value = edtf_date(input_date)
+    end
+
+    def edtf_date(date)
+      edtf_val = Date.edtf(date)
+      if edtf_val.blank? && date.present?
+        # date_uploaded usually comes in the form m/d/Y unlike edtf that comes in the form Y-m-d
+        custom_date = Date.strptime(date, '%m/%d/%Y')
+        edtf_val = Date.edtf(custom_date.to_s)
+      end
+      edtf_val
+    rescue StandardError => e
+      Rails.logger.warn "GoogleScholarCitationService #{e.message}"
+      nil
+    end
+
+    def format_date
+      if @edtf_value.instance_of? EDTF::Interval
+        @edtf_value.to_s
+      elsif @edtf_value.present?
+        date_from_edtf(@edtf_value)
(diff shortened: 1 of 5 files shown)


luisgreg99:
Update for checklist item 3: @simholt already took care of this special case (date intervals). Thank you Sarah!

app/services/scholars_archive/google_scholar_citation_service.rb
```diff
@@ -0,0 +1,72 @@
...
+
+module ScholarsArchive
+  # Class to handle the expected format for citation publication dates as
+  # required in the indexing guidelines for configuring meta-tags:
+  # https://scholar.google.com/intl/en/scholar/inclusion.html#indexing
+  class GoogleScholarCitationService
+    attr_reader :input_date
+
+    def self.call(input_date)
+      new(input_date).call
```
straleyb: I think you could cut out the extra #call here. `new(input_date).format_date` would work peachy

app/services/scholars_archive/google_scholar_citation_service.rb
```diff
@@ -0,0 +1,72 @@
...
+      edtf_val = Date.edtf(date)
+      if edtf_val.blank? && date.present?
+        # date_uploaded usually comes in the form m/d/Y unlike edtf that comes in the form Y-m-d
+        custom_date = Date.strptime(date, '%m/%d/%Y')
+        edtf_val = Date.edtf(custom_date.to_s)
+      end
+      edtf_val
+    rescue StandardError => e
+      Rails.logger.warn "GoogleScholarCitationService #{e.message}"
+      nil
```
straleyb: Is this improperly indented?

app/services/scholars_archive/google_scholar_citation_service.rb
luisgreg99: good call

app/services/scholars_archive/google_scholar_citation_service.rb
luisgreg99: It's intended to catch errors like invalid inputs (and return `nil` so that if a date is invalid, it may proceed with an empty value without breaking the show page). `Date` may fail with `ArgumentError: invalid date`. See spec https://github.com/osulp/Scholars-Archive/blob/google-scholar-citation-date/spec/services/scholars_archive/google_scholar_citation_service_spec.rb#L19

PR #1907, opened 2019-07-23
bump ruby and update gems

straleyb:
This bumps ruby to 2.5.5 and updates gems to work. I disabled clamav though because it doesnt work in docker.


Dockerfile
```diff
@@ -12,11 +12,11 @@ RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - && \
 RUN gem install bundler
 
 RUN apt-get update -qq && apt-get upgrade -y && \
-  apt-get install -y build-essential libpq-dev mysql-client nodejs libreoffice imagemagick unzip ghostscript yarn
+  apt-get install -y build-essential libpq-dev mariadb-client nodejs libreoffice imagemagick unzip ghostscript yarn
 
 # install clamav for antivirus
 # fetch clamav local database
-RUN apt-get install -y clamav-freshclam clamav-daemon libclamav-dev
+RUN apt-get install -y clamav-freshclam clamav-daemon libclamav-dev clamav-base
```
wickr: If we're disabling clamav for Docker we can comment out or remove these lines, it'd speed up Docker installs.

PR #1929, opened 2019-08-28
updates popup

### LINKED ISSUE #1557: Configure multiple tooltips depending on the context of the link.
### Descriptive summary

On the show page, we need to parse out the specific links that exist in the metadata of a work. If the link contains an @ then append a specific class so the JavaScript can enable the proper tool tip. Otherwise, check the parent field, if it is a description, then it should say "visit external link", finally, the rest should say "search SA@OSU"

### Expected behavior

Popup helper text should say "Visit external link" or something to that effect.

### Actual behavior

Popup helper text says "Search for this term in SA@OSU"

### Steps to reproduce the behavior

1. Visit https://ir.library.oregonstate.edu/concern/graduate_thesis_or_dissertations/kd17ct94m?locale=en
2. Look at Additional Information section, hover over email address link

### Screenshot
image

straleyb:
Fixes #1557

### DIFF
--- a/app/assets/javascripts/tooltips.js
+++ b/app/assets/javascripts/tooltips.js
@@ -1,7 +1,7 @@
 Blacklight.onLoad(function() {
   var search_options = {
     placement: 'top',
-    title: 'Search for this term in SA@OSU'
+    title: 'Visit external link'
   };
   var more_information = {
     placement: 'top',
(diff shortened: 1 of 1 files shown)


simholt:
@wickr Having a mailto one sounds like a good idea. Most of the email addresses in the system are from the approval process (see below) but some depositors do list addresses for other reasons.

Screen Shot 2019-09-04 at 11 41 24 AM

PR #1930, opened 2019-08-28
update config schema

luisgreg99:
Update solr config schema to address solr exception on fresh docker build:
```
RSolr::Error::Http - 500 Internal Server Error Error: { "error":{ "metadata":[ "error-class","org.apache.solr.core.SolrCoreInitializationException", "root-error-class","java.lang.ClassNotFoundException"], "msg":"SolrCore 'hydra-development' is not available due to init failure: 
```

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -48,13 +48,16 @@
 <schema name="Hydra Demo Index" version="1.5">
   <!-- attribute "name" is the name of this schema and is only used for display purposes.
        Applications should change this to reflect the nature of the search collection.
-       version="1.4" is Solr's version number for the schema syntax and semantics.  It should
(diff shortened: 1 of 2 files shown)


straleyb:
Ive ran `docker system prune` and `docker-compose build` multiple times, and I cant seem to get this error to pop up. Can you write steps to reproduce this error?

wickr:
@straleyb 	did you have luck doing `docker-compose up server` ?  docker-compose build is fine, but Solr has issues on startup due to the conf files

wickr:
I rebuilt again and Solr seems to start up within Docker without issues. (though I can't create a work yet to check, workflows don't seem to want to load)

straleyb:
Stale, If we run into this issue again, we can revisit this.

PR #1959, opened 2019-09-19
handle link is external now

### LINKED ISSUE #1152: Change DOIs so they do not search internally when clicked
DOI links should be external, not internal - ie submitting a search query, which is not helpful.

straleyb:
Fixes #1152

### DIFF
--- a/.rubocop_fix_me.yml
+++ b/.rubocop_fix_me.yml
@@ -1119,6 +1119,7 @@ RSpec/ExampleLength:
     - 'spec/actors/scholars_archive/actors/nested_fields_operations_actor_spec.rb'
     - 'spec/actors/scholars_archive/actors/add_other_field_option_actor_spec.rb'
     - 'spec/inputs/multi_value_ordered_input_spec.rb'
+    - 'spec/renderers/hyrax/renderers/external_link_attribute_renderer_spec.rb'
 
 RSpec/MultipleExpectations:
   Exclude:
--- a/app/renderers/hyrax/renderers/external_link_attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/external_link_attribute_renderer.rb
@@ -8,7 +8,7 @@ module Hyrax
 
         def li_value(value)
           normalized_link = value.include?('http://') ? auto_link(value) : "http://#{auto_link(value)}"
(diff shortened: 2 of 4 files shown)


straleyb:
Screen Shot 2019-11-06 at 2 34 49 PM
Screen Shot 2019-11-06 at 2 35 53 PM

luisgreg99:
@straleyb It looks like this update wound't work if the `doi` uses the `https` protocol like in this work https://ir.library.oregonstate.edu/concern/articles/1n79h487k where `doi` is:
```
https://doi.org/10.1577/1548-8446(2007)32[540:DOASDD]2.0.CO;2
```

I would recommend this method `maybe_uri(s)` available in `search_and_external_link_attribute_renderer`. See https://github.com/osulp/Scholars-Archive/blob/master/app/renderers/hyrax/renderers/search_and_external_link_attribute_renderer.rb#L51  

`maybe_uri` takes care of `http` and `https` protocols, including special characters in the `doi` like `(`,`)`, `[`, `]` in the example https://ir.library.oregonstate.edu/concern/articles/1n79h487k.

You could copy `maybe_uri(s)` in your renderer, but I think it would be better to inherit from `SearchAndExternalLinkAttributeRenderer` instead of `AttributeRenderer`. This line https://github.com/osulp/Scholars-Archive/blob/feature/HDLExternalLink/app/renderers/hyrax/renderers/external_link_attribute_renderer.rb#L6 could be 
```
class ExternalLinkAttributeRenderer < SearchAndExternalLinkAttributeRenderer
```
and then `maybe_uri(s)` should be available in the class `ExternalLinkAttributeRenderer`.

luisgreg99:
@straleyb Ok, it looks like the urls with `https` are now working. I tested it locally, and it looks good. 

My only observation (now that I had the chance to test it locally) is the "new window" icon `glyphicon-new-window`. I think for consistency with other fields, we should remove it or maybe link it with `target="_blank"` so that it behaves like the other links (it should open the link on a new tab/window).

For example, in `Right statements`, the icon `glyphicon-new-window` opens a link on a new tab, but in `Doi` it doesn't do anything (unless you click on the actual link), so it can be confusing:

image

Issue #1977, opened 2019-10-31
Revise Ack message 


shieldsb:
### Descriptive summary
Revise Ack message 

### Expected behavior
Message that explains SA outage

simholt:
Please change message to:

> Something isn’t working on our end. Please try again.
> If you encounter this message repeatedly, please let us know what happened via the [contact form](https://ir.library.oregonstate.edu/contact).

Link to contact form: https://ir.library.oregonstate.edu/contact

simholt:
QA passed.

PR #1978, opened 2019-11-04
update to hyrax 2.6

straleyb:


luisgreg99:
@straleyb I think for this PR, it would be a good idea to address this ticket first https://github.com/osulp/Scholars-Archive/issues/1977 (Revise Ack message) before merging and deploying to production just in case we get 500 server errors. I think having a more user friendly message would be nice for the unexpected errors due to software upgrades.

PR #1983, opened 2019-11-06
sort based on metadata name

### LINKED ISSUE #1981: Alpha sort advanced search fields
### Descriptive summary

The advanced search page contains multiple fields that users can search against. These fields need to be alpha sorted. Currently they have no logical order.

straleyb:
Fixes #1981


luisgreg99:
@straleyb can you add a screenshot and a note in the partial about your changes? It looks like you got the base file from blacklight gem right? I think an override tag would be good https://github.com/projectblacklight/blacklight_advanced_search/blob/master/app/views/advanced/_advanced_search_fields.html.erb

straleyb:
Screen Shot 2019-11-12 at 9 53 22 AM

PR #2009, opened 2019-12-06
CircleCI refactor PR

decimalator:
Refactor CircleCI workflow configuration to look more like OD2, which looks more like SA in production with external services in their own containers.

Also pushes built containers to the Beavernetes registry to facilitate Beavernetes deployment


solr/conf/schema.xml
```diff
@@ -329,11 +329,17 @@
...
+ <!-- Field to use to determine and enforce document uniqueness. 
       Unless this field is marked with required="false", it will be a required field
    -->
  <uniqueKey>id</uniqueKey>
 
+ <!-- field for the QueryParser to use when an explicit fieldname is absent -->
+ <!--  <defaultSearchField>text</defaultSearchField> -->
+
+ <!-- SolrQueryParser configuration: defaultOperator="AND|OR" -->
+ <solrQueryParser defaultOperator="OR"/>
```
CGillen: Specifically what happened here?

Dockerfile
```diff
@@ -34,12 +32,11 @@ ADD Gemfile /data/Gemfile
 ADD Gemfile.lock /data/Gemfile.lock
 RUN mkdir /data/build
 
-ARG RAILS_ENV=development
+#ARG RAILS_ENV=development
```
straleyb: If this is commented out, will we need to keep this?

straleyb:
Looks good. Im confused about why some files say their content was deleted then added, but the contents look the same. But if this just has to do with CCI i'd call it good once the two comments are addressed and we can get this in.

solr/conf/schema.xml
decimalator: these are the live configs I merged from prod when I was trying to figure out which set of configs was actually being used since there's both a `solr/conf` and a `solr/config`. So this gets us closer to how SOLR should be running, in theory

Dockerfile
decimalator: it's not _needed_ per se. I commented it out because at one point it was running in `development` mode in CCI, so I unset it in the `Dockerfile` in case that was why it wasn't obeying the value being set in the circle config.yml

Everywhere that this runs (docker-compose for dev, the `monit` infra for staging/prod) is forcing the RAILS_ENV value to however it's being run. 

I would rather have to explicitly set `RAILS_ENV` in deployments and have it fail if you don't set it than have it set a default that might be wrong and potentially cause weird behavior

PR #2020, opened 2020-01-17
[WIP] First step to integrating fetch in job

straleyb:
Needs tests ported over and added to all models.


straleyb:
Redoing to pull new work from OD2

PR #2037, opened 2020-02-19
Tombstone mechanism in place

### LINKED ISSUE #1801: Create tombstone pages for removed items
### Descriptive summary
As part of our promise of persistency for works in SA@OSU, we need to implement tombstone pages for works that have been removed from the repository. This should be a page that the URI for the item resolves to with text to indicate that the item in question has been removed or withdrawn from the system.

straleyb:
Fixes #1801 

This sets a work as tombstoned and private. We can check the state of works and see that they are tombstoned. They are also set to private so they are undiscoverable to anyone but administrators of collections.


straleyb:
tank

PR #2042, opened 2020-02-20
update to use date issued instead of date created

### LINKED ISSUE #818: Set up and configure auto generation of bibliographic citations
Hyrax has the ability to generate MLA/APA/Chicago citations automatically for items, but it needs to be set up and surfaced.

straleyb:
Fixes #818 

Sets up citations to use date issued


straleyb:
Borked. Rebuilding this PR

PR #2061, opened 2020-03-02
Feature/form updates

straleyb:


app/models/concerns/scholars_archive/article_metadata.rb
```diff
@@ -32,6 +33,50 @@ module ArticleMetadata
       property :web_of_science_uid, predicate: ::RDF::URI.new('http://opaquenamespace.org/ns/webOfScienceUid'), multiple: false do |index|
         index.as :stored_searchable
       end
+
+      define_singleton_method :article_properties do
```
CGillen: Do you have plans for these methods? They don't seem used in this PR and didn't exist before as far as I can tell.

straleyb: Yeah, I wanna get these initial refactors in and get it all playing nicely. This method would replace the model::TERMS which should lower the amount of lists we carry around smaller.

config/locales/hyrax.en.yml
```diff
@@ -85,7 +85,7 @@ en:
 
     works:
       create:
-        after_create_html: "ScholarsArchive@OSU has received your upload. Your file is currently being processed. Thank you for your deposit."
+        after_create_html: "ScholarsArchive@OSU has received your deposit. Your file is currently being processed; some fields may take a few minutes to display."
```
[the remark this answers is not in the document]
straleyb: Yeah..... Its git funny business.

straleyb:
@CGillen Yeah, I can do that. Ill move everything into their own modules. I think thats the ultimate goal here.

PR #2070, opened 2020-03-09
Feature/file set delete ldp gone tombstone

### LINKED ISSUE #1801: Create tombstone pages for removed items

straleyb:
Fixes #1801

Still need to implement javascript popup for "Are you sure?" for tombstone process.

### DIFF
--- a/app/views/hyrax/base/_show_actions.html.erb
+++ b/app/views/hyrax/base/_show_actions.html.erb
@@ -11,7 +11,7 @@
       <% if @presenter.workflow.actions.present? && @presenter.solr_document['workflow_state_name_ssim'].first == 'Deposited' %>
         <%= form_tag main_app.hyrax_workflow_action_path(presenter), method: :put do %>
           <%= hidden_field :workflow_action, :name, value: 'tombstone' %>
-          <input class="btn btn-danger" type="submit" value="Tombstone" onclick='confirm("Are you sure you would like to tombstone this work?");'>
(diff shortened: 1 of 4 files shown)


app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -8,10 +8,10 @@
     <%# Changes start here %>
     <% if current_user.admin? %>
       <%= link_to "Delete", [main_app, presenter], class: 'btn btn-danger', data: { confirm: "Delete this #{presenter.human_readable_type}?" }, method: :delete %>
-      <% if @presenter.workflow.actions.present? && @presenter.solr_document['workflow_state_name_ssim'].first == 'deposited' %>
+      <% if @presenter.workflow.actions.present? && @presenter.solr_document['workflow_state_name_ssim'].first == 'Deposited' %>
         <%= form_tag main_app.hyrax_workflow_action_path(presenter), method: :put do %>
           <%= hidden_field :workflow_action, :name, value: 'tombstone' %>
-          <input class="btn btn-danger" type="submit" value="Tombstone">
+          <input class="btn btn-danger" type="submit" value="Tombstone" onclick='confirm("Are you sure you would like to tombstone this work?");'>
```
luisgreg99: I think it would be a good idea to set the confirmation message in the en locale file

PR #2141, opened 2020-10-05
update triplestore-adapter, hyrax, and rails

straleyb:


db/migrate/20170317142609_create_searches.blacklight.rb
```diff
@@ -1,6 +1,6 @@
 # This migration comes from blacklight (originally 20140202020201)
 # frozen_string_literal: true
-class CreateSearches < ActiveRecord::Migration
+class CreateSearches < ActiveRecord::Migration[4.2]
```
CGillen, line 3: Maybe you know more about this, but from what I can find the number here is the rails version target? Should these be `5.2` or did something tell you `4.2` is right?

db/schema.rb
```diff
@@ -10,571 +10,571 @@
...
-    t.datetime "created_at",             null: false
-    t.datetime "updated_at",             null: false
-    t.index ["order"], name: "index_featured_works_on_order", using: :btree
-    t.index ["work_id"], name: "index_featured_works_on_work_id", using: :btree
-  end
-
-  create_table "file_download_stats", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1" do |t|
+ActiveRecord::Schema.define(version: 2018_05_29_220525) do
+
+  create_table "bookmarks", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
```
CGillen, line 15: I'm wondering how db/schema.rb got so many changes, particularly the `CHARSET` it seems dangerous to just suddenly have our schema dump switching character sets if our database hasn't gone through a proper migration.
`config/database.yml` and `config/mysql/mysqld.cnf` seem to disagree on this as well. This might be good to double check with the live database to see what character set it's using now. imo we should be using utf8, but I'd rather not corrupt data.

config/initializers/new_framework_defaults.rb
```diff
@@ -18,7 +18,7 @@
 Rails.application.config.active_record.belongs_to_required_by_default = true
 
 # Do not halt callback chains when a callback returns false. Previous versions had true.
-ActiveSupport.halt_callback_chains_on_return_false = false
+# ActiveSupport.halt_callback_chains_on_return_false = false
```
CGillen, line 21: Making a note here, supposedly this has become default behavior in rails 5, so rails 5.2 removed the option. Nothing should have changed by removing this.

db/migrate/20170317142609_create_searches.blacklight.rb
straleyb: I tried 5.2 but there are other issues that arise from that. All the other migrations in the app emulate the 4.2 change, so its actually more consistent to have it this way specifically for hyrax.

db/schema.rb
straleyb: Yeah im not entirely sure where this change came, but i assume its part of the reason why i had to drop the database and rebuild the schema. I can look deeper.

config/initializers/new_framework_defaults.rb
straleyb: Yep. This actually moved out of active support, so it exists elsewhere. So I commented it out to keep it around just in case something weird happens, but it should exist elsewhere still.

PR #2145, opened 2020-10-29
update workflows page to remove published

### LINKED ISSUE #1133: load times for Review Submissions page is too slow
wait time of up to 15 seconds is too long. please check to see if performance can be enhanced! Do not show published tab.

straleyb:
Fixes #1133


CGillen:
@simholt It's been a while since we've looked at this. Is the review queue loading time still an issue and is removing the `published` tab still an acceptable solution?

simholt:
@CGillen Slow loading is occasional but I would still like the `published` tab removed, yes!

CGillen (approved):

PR #2182, opened 2021-02-17
Replace sorting fields with system created dates and non-ordered titles

### LINKED ISSUE #667: Add 'Sort by Title' options to search results
Add 'Sort by Title A-Z' and 'Sort by Title Z-A' as options in the search results. May need to figure out what exactly in the title will be used for the sort.

CGillen:
`title_ssi` is populated with the first ordered or unordered title, so that one should be reasonably similar behavior.
The system fields are generated by the actor stack and guaranteed to be on works, unlike `date_created` which is an optional field. `date_uploaded` is only applied to works, so collections get miss-sorted.

Reverts and Fixes #667

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -13,7 +13,7 @@ class CatalogController < ApplicationController
   before_action :enforce_show_permissions, only: :show
 
   def self.uploaded_field
-    solr_name('system_create', :stored_sortable, type: :date)
+    solr_name('date_uploaded', :stored_sortable, type: :date)
   end
 
   def self.title_field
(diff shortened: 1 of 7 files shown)


CGillen:
Ok, reworked as follows:
`title_ssi` is used for title sort. This value should still be the first of ordered/unordered titles

`date_uploaded_dtsi` is still used for date uploaded sort. Now if there's no value in date fields those docs get pushed to the bottom by solr's `sortMissingLast` property

A new field is generated on works and collections, `date_sort_combined_dtsi`. This is generated by picking the first of:
`date_issued date_created date_copyright`. Again, pushed to the bottom if none of those are filled out. `date_created` is the only one that applies to collections but the logic uses safe access and will treat the other fields as nil instead of throwing an exception.

wickr (approved):

PR #2218, opened 2021-03-30
Update rails to remove mimemagic dependency.

wickr:
Mimemagic gems were pulled, and though newer versions were released, it seems the best path forward is to update Rails which no longer relies on it.

https://github.com/rails/rails/issues/41750

Some of the samvera/hydra gems and honeycomb were updated as well.


wickr:
Closing in favor of #2219

Issue #1809, opened 2019-02-14
Diacritics and sorting results


vantuyls:
### Descriptive summary

we would lke to diacritics to sort alphabetically with their assocaited alphanumeric letters (e.g. u sorted next to a ü).

What are our sorting options/configurations for such things in SOLR/Hyrax?

CGillen:
#2212 prepped the indexers to index into `title_ssort` which ignores diacritics and upper/lowercase as a more natural sorting method. Once `title_ssort` is populated [these lines](https://github.com/osulp/Scholars-Archive/pull/2212/files#diff-bd4eb77984b740347ff2aa902be664d1aa01addc73964897450d5bd3ff09b3c6R614-R617) can be used in the catalog controller in place of the title sort directly above them.

CGillen:
Fixed by #2239

simholt:
The title sorts look great!
I'm realizing this ticket was not specific -- can we set this up for all fields where names are used? That's where most diacritic words live.

- advisor
- creator
- contributor
- journal title (sounds like a good idea)

CGillen:
@simholt Do I need to expose these sorts anywhere on the website? I can see how these would be helpful for your backend solr queries, but I want to make sure that indexing is the only thing that needs to be added

simholt:
@CGillen My guess is an index would enable the type of sort shown below? Diacritics currently appear at the end when users pop out the limiters (see below) and use the A-Z Sort setting. Although I remember talking with @wickr previously and I don't think Contributors are currently sortable (another issue).

![Screen Shot 2021-05-19 at 2 40 09 PM](https://user-images.githubusercontent.com/33705514/118888159-25a11280-b8b0-11eb-8777-ef163a822da6.png)

CGillen:
Ok, that can be done but can those facets be all lowercase?

The way the sorting is indexed is by lowercasing and folding any non-latin characters into the most compatible latin character. Normally solr would also store the original for display in the system, but the facet system in particular can only retrieve the index. So we're forced into using the normalized version.

simholt:
Yes that sounds good!

CGillen:
#2260 gets the indexer ready to support new `*_ssort` fields for these facets:

-    advisor
-     creator
-     contributor
-     journal title
Once that is on prod and a reindex occurs, the notes from #2260 can be implemented and the facets turned on

simholt:
QA pass

PR #2261, opened 2021-05-21
Add new solr field for alphasorted facets

CGillen:
Reconfigure catalog controller and work indexer for new *_sfacet fields


CGillen:
@rordway Do we have a way to deploy changes to the solr schema? My initial take on this used a field that wasn't multivalued and now I need to add a new dynamicfield that is multivalue. If we have a reasonable way to do that, this PR can go through so that the reindex needed for #1809 can be done

PR #2280, opened 2021-07-21
add SMTP settings

decimalator:


CGillen (changes requested):
Make sure these (and all the other) environment variables exist on production

decimalator:
> Make sure these (and all the other) environment variables exist on production

all added environment variables were added to the production local_env.yml

Issue #1782, opened 2018-12-26
Replace conference_section predicate on all works needed


wickr:
### Descriptive summary

Once #1781 is completed, find all works with the old, wrong predicate for conference_section: `https://w2id.org/scholarlydata/ontology/conference-ontology.owl#Track`

Can't rely on just Solr.

Replace with correct predicate on those works and save.

### Related work

#1781

straleyb:
Relates #1861

CGillen:
Rake task ran, works updated

PR #2310, opened 2021-11-29
Update conference section predicate

### LINKED ISSUE #1781: Fix predicate for :conference_section
### Descriptive summary

The predicate for `conference_section` got changed and needs to be corrected (`...w2id.org...` should be `...w3id.org...`)

Not sure how it got changed, but it was by me when the conference predicates were moved to default metadata: https://github.com/osulp/Scholars-Archive/commit/c7f60452509ca3184cd1111715e67eadfd2d5c91#diff-1e841b2558a016e0f500e5866610cb70R42

Works with the old predicate will not have that data show up if edited and saved or reindexed.

### Expected behavior

`conference_section` predicate is `https://w3id.org/scholarlydata/ontology/conference-ontology.owl#Track`

### Actual behavior
`conference_section` predicate is `https://w2id.org/scholarlydata/ontology/conference-ontology.owl#Track`

### Related work

#1782 - Replace predicate on all works needed

CGillen:
This will update the conference section predicate to use `w3id` rather than `w2id`. After merging we will want to quickly run `rake scholars_archive:update_conference_section_data` to update all the old values on all works to the new predicate.

supersedes #2282
Fixes #1781, #1782

### DIFF
--- a/app/jobs/replace_predicate_job.rb
+++ b/app/jobs/replace_predicate_job.rb
@@ -13,33 +13,25 @@ class ReplacePredicateJob < ScholarsArchive::ApplicationJob
     logger.info "Updating #{ids_and_value.count} works"
     logger.info ids_and_value
     ids_and_value.each do |id, value|
-      logger.info value
-      logger.info id
       work = ActiveFedora::Base.find(id)
-      logger.info "Work Found"
       old_predicate_statement = [work.resource.rdf_subject, ::RDF::URI.new(old_predicate), nil]
-      logger.info "Predicate set"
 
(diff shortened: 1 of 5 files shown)


CGillen:
Wait until things settle down with the Fedora migration and get the go-ahead from @simholt. This shouldn't have an impact on the user experience. Front-end display is based on Solr, which won't change. Back-end, we'll still be able to ingest new works without problem. Editing works will still work if they save before the work is reached by the update job, or they open the edit after it's reached. **A long standing edit page could be a problem for works with a conference section.** Because of this we may want to start near the end of the work day or week

PR #2353, opened 2022-03-04
remove capistrano stuff

straleyb:


CGillen (changes requested):
https://github.com/osulp/Scholars-Archive/blob/master/.dockerignore#L9
https://github.com/osulp/Scholars-Archive/blob/master/Dockerfile#L72
https://github.com/osulp/Scholars-Archive/blob/master/bin/cap
https://github.com/osulp/Scholars-Archive/blob/master/bin/capify

Just a few more small spots

CGillen (approved):

PR #2408, opened 2022-10-25
Add Core Trust Logo to Footer

lamtu1:
Adding the Core Trust Seal logo to the footer of the Scholars Archive and align it to match the footer layout.


docker-compose.yml
```diff
@@ -130,6 +130,7 @@ services:
   # Fedora repository server
   blazegraph-image: &blazegraph-image
     image: lyrasis/blazegraph:2.1.5
+    platform: linux/amd64
```
CGillen, line 133: Can you uncommit this file? I'm not sure this fix actually works well enough to commit yet. We might have to do something similar to OD2, where we rebuild some images

app/views/shared/_footer.html.erb
```diff
@@ -11,6 +11,13 @@
         <p><%= t('hyrax.footer.copyright_html') %></p>
         <p class='deployed'><%= render partial: 'shared/application_revision' %></p>
       </div>
+
+      <%# NAVBAR-IMAGE: Add in the Core Trust Seal onto Scholars Archive %>
+      <div class="navbar-image">
+        <%= link_to "https://www.coretrustseal.org/wp-content/uploads/2022/09/20220922-ScholarsArchive@OSU_final.pdf" do %>
+          <%= image_tag("CoreTrustSeal-logo-transparent.png", size: "80x80", alt: "Core Trust Seal Logo (PDF)") %>
```
CGillen, line 18: @straleyb I think this is an odd a11y edge case. We have a link with a logo that's text (alt text should include text & should specify it's a logo), it also links out (should have text in the link that tells user about where they're going), and the link goes to a pdf (should tell user they're leaving to a file, especially a pdf, that's why I had Tu add `(PDF)` to the alt text).

What do you think here? Maybe we should add a `.sr-only` bit of text below the image that says something like `Core Trust application (pdf)` and drop the `(PDF)` from the alt text?

CGillen (changes requested):

app/views/shared/_footer.html.erb
straleyb: Yeah I think having an `.sr-only` is probably the better option here. Since its a little more complex, we can give the person a little more of an idea as to whats going on. We should keep the alt text obviously, but having a screen reader only seems to make a lot of sense here.

straleyb: Also, im not sure if this is necessary but maybe we should move the size out of the image tag and use a class/css instead.

lamtu1: I can move that size dimension to it owns class and have it there and not have it clutter the image tag with many options and keeping the alt tag there only.

straleyb (approved):

CGillen:
Congrats @lamtu1 on your first production PR merge 🎉 🎉

PR #2409, opened 2022-10-26
add a HONEYCOMB_ENABLED toggle 

decimalator:
For local dev and to allow us to easily disable Honeycomb output

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,10 +1,14 @@
 # Honeycomb integration
 
-if Rails.env.development? || ENV.key?("HONEYCOMB_DEBUG")
-  Honeycomb.configure do |config|
-    config.client = Libhoney::NullClient.new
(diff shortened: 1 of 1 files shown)


CGillen (approved):

config/initializers/honeycomb.rb
```diff
@@ -34,4 +38,6 @@
       ScholarsArchive::NoiseCancellingSampler.sample(fields)
     end
   end
+else
+   Rails.logger.warn 'Honeycomb disabled, skipping initialization'
```
CGillen, line 42: @decimalator We need to do some amount of configuration when `HONEYCOMB_ENABLE` is disabled, otherwise `#append_info_to_payload` gets upset. It can be as simple as
```
  Honeycomb.configure do |config|
  end
```
but idk if there's any other configuration you need to do to silence Honeycomb

PR #2412, opened 2022-11-02
Prepopulate the license only in Dataset

lamtu1:
Adding in a new file to override the base license file to have it populate the copyright in the box in the dataset section only.


straleyb:
Looks good to me. I'd like to see a screenshot of the field being preselected for datasets but not for another. Otherwise, this looks good on my end.

lamtu1:
https://user-images.githubusercontent.com/111306182/199578803-8602c06a-5072-40fa-a8d4-06ec53abeb3d.mp4

Rather than a screenshot, I did a short demo showing the result, so it would make it clear that the change did happen.

straleyb (approved):

PR #2418, opened 2022-11-07
Update timedate format based on feedback

lamtu1:
Given feedback on the datetime, we formatted a way that differentiate between which comments belong under which comment.

### DIFF
--- a/app/assets/stylesheets/osu.scss
+++ b/app/assets/stylesheets/osu.scss
@@ -523,6 +523,11 @@ ol.dd-list.dd-dragel {
   color: #808080;
 }
 
+.comment-line {
+  margin-top: -10px;
+  margin-bottom: 20px;
+}
+
 textarea.nested-field {
   min-height: 68px !important;
 }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/osu.scss
```diff
@@ -523,6 +523,11 @@ ol.dd-list.dd-dragel {
   color: #808080;
 }
 
+.comment-line {
+  margin-top: -10px;
```
CGillen, line 527: Instead of using a negative margin, try using a 0 margin on the `.comment-time` class

CGillen (changes requested):

app/assets/stylesheets/osu.scss
lamtu1: Screenshot 2022-11-07 at 11 18 29 AM

If I didn't add in the -10px, this is what the 0 margin-top does to the comment.

CGillen:
```
.comment-time {
  font-size: 10px;
  font-style: italic;
  color: #808080;
  ~> margin-bottom: 0; <~
}
```

lamtu1: Ah I see, adding the 0px to the timedate for it to push up the comment, I will push in that new update now.

CGillen (approved):

PR #2430, opened 2022-12-08
Append summary log onto Fixity log

lamtu1:
Adding in couple report to the log output using `ChecksumAuditLog`.


lib/tasks/fixity_check.rake
```diff
@@ -2,9 +2,75 @@ namespace :scholars_archive do
...
+    end_time = Time.now
+
+    # QUERY #1: Query data from the ChecksumAuditLog
+    latest_file = ChecksumAuditLog.where("updated_at >= ?", start_time)
+    file_checked = latest_file.count
+    file_passed = latest_file.where("passed = true").count
+    file_failed = latest_file.where("passed = false").count
+
+    # QUERY #2: Get all the ids that failed via checking with fixity
+    latest_file.each do |c|
```
CGillen, line 30: You already have all the failed files here: `latest_file.where("passed = false")`
Then you can use a `Array#map` method to extract & capture the `file_set_id`

```
failed_arr = latest_file.where("passed = false").map(&:file_set_id)
```

lib/tasks/fixity_check.rake
```diff
@@ -2,9 +2,75 @@ namespace :scholars_archive do
...
+    Rails.logger.info "'No. of File Sets [CHECKED]': #{file_checked.to_s}"
+    Rails.logger.info "'No. of File Sets [PASSED]': #{file_passed.to_s}"
+    Rails.logger.info "'No. of File Sets [FAILED]': #{file_failed.to_s}\n"
+
+    Rails.logger.info "'No. of File Sets [IDs] that failed':"
+    if (failed_arr.empty?)
+      Rails.logger.info "None\n"
+    else
+      failed_arr.each do |f|
+        Rails.logger.info "#{f.to_s}"
```
CGillen, line 51: Ask @decimalator if he wants a failed ID per line, or a single summary array line.

spec/mailers/scholars_archive/fixity_mailer_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      num_file: 21,
+      file_pass: 18,
+      file_fail: 3,
+      fail_arr: %w[1 2 3] }
+  end
+  let(:mail) { described_class.with(user: user, data: tst_data).report_email }
+
+  # TEST GROUP: Create couple test to see if the mailer class pass the test
+  it { expect(mail.subject).to eql('Scholars Archive: Fixity Report') }
+  it { expect(mail.to).to eql([user.email]) }
```
CGillen, line 19: I thought of some tests you could add
Make sure the mail has each of these lines with the right values from `tst_data`
```
    <p>Start Time: <%= @fixity_data[:start_time] %></p>
    <p>End Time: <%= @fixity_data[:end_time] %></p>
    <p>No. of Files [CHECKED]: <%= @fixity_data[:num_file] %></p>
    <p>No. of Files [PASSED]: <%= @fixity_data[:file_pass] %></p>
    <p>No. of Files [FAILED]: <%= @fixity_data[:file_fail] %></p>
```

CGillen (changes requested):

lib/tasks/fixity_check.rake
decimalator: I think let's try a single summary array and see how that looks. 

as long as it's one line it can be as long as we want it to be. I think there's a limit to the length of a single log push to  Loki, I'll verify that. but the summary array shouldn't be any longer than some of the logs I've seen come through

lamtu1: Hi @decimalator, below is two screenshots if we printed them out onto the log console via each `individual item` or in a `single summary array` format. Let me know if you would still like to continue with the `single summary array` style after checking them out below.

PS: This is all test data to display

`Option #1`: With `individual item`
Screenshot 2022-12-14 at 10 24 40 AM

`Option #2`: With the `single summary array`
Screenshot 2022-12-14 at 10 36 09 AM

decimalator: I like option 2! looks great!

lamtu1: @decimalator, thank you for the feedback, I will go with it. Also, another question I wanted to ask, do you want the single summary array be printed in the email format too or keep them original format right now which follow the resemblance of option 1?

CGillen (approved):

Issue #1013, opened 2017-11-14
OAI, connect Set description to Admin Set abstract


wickr:
Currently there's no OAI Set Description configured (just title). This isn't critical, but some harvesters may display the description somewhere in their interface.

This can pull from the Admin Set abstract/summary, though currently none of them have any values there.

lamtu1:
Fixed by #2435

PR #2435, opened 2023-01-04
Connect description onto OAI

lamtu1:
Adding the new field `SetDescription` on to the `OAI`. This can be useful to have when it comes to get more data out of the metadata in `OAI`.

### DIFF
--- a/app/models/oai_set.rb
+++ b/app/models/oai_set.rb
@@ -37,12 +37,25 @@ class OaiSet < BlacklightOaiProvider::SolrSet
   def initialize(spec, opts = {})
     @spec = spec
     @name = opts[:name] || name_from_spec
-    @description = opts[:description]
+    @description = opts[:description] || description_from_spec
   end
 
   private
 
   def name_from_spec
     ActiveFedora::SolrService.query("has_model_ssim:AdminSet AND id:#{@spec}", rows: 1).first['title_tesim'].first
+  end
+
+  # METHOD: Add in an option to fetch the metadata description to add onto OAI
+  def description_from_spec
+    # QUERY: Use the query to get the description per collection using save navigation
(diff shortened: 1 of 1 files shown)


lamtu1:
Below is the result of what the `code` will display on the `OAI`:
Screenshot 2023-01-04 at 10 57 09 AM

app/models/oai_set.rb
```diff
@@ -37,12 +37,21 @@ def sets_from_facets(facets)
...
     ActiveFedora::SolrService.query("has_model_ssim:AdminSet AND id:#{@spec}", rows: 1).first['title_tesim'].first
   end
+
+  # METHOD: Add in an option to fetch the metadata description to add onto OAI
+  def description_from_spec
+    # QUERY: Use the query to get the description per collection using save navigation
+    describe_meta = ActiveFedora::SolrService.query("has_model_ssim:AdminSet AND id:#{@spec}", rows: 1).first['description_tesim']&.first
+
+    # CONDITION: Check the condition to return the correct description to display on OAI
+    return describe_meta if describe_meta != '' || describe_meta.nil?
```
CGillen, line 55: Check the functionality of `#blank?` or `#empty?`. I'm not sure which works better in this case

CGillen (changes requested):

app/models/oai_set.rb
lamtu1: The functionality `blank?` would work and work on `nil` as well but not for `empty?` but I'm currently writing in a safe guard statement due to RuboCop hate if I write if-else block. But when i tried `blank?` only the one with the `[""]` will print out and nothing else it is print out.

lamtu1: Got it fixed and pushing it as a PR.

lamtu1:
PR Fixes #1013

CGillen (approved):

PR #2453, opened 2023-04-11
Add in bug fix to the delete all button

lamtu1:
Fix the bug where the button is still clickable when it is disable, and change to present if FileSets are presented there

### DIFF
--- a/app/views/hyrax/base/_items.html.erb
+++ b/app/views/hyrax/base/_items.html.erb
@@ -5,7 +5,7 @@
 
 <%# BUTTON: Add in the deletion of all files %>
 <div class="btn-line">
-  <% if presenter.file_set_presenters.blank? == false %>
+  <% if !presenter.file_set_presenters.blank? %>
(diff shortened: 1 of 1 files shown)


app/views/hyrax/base/_items.html.erb
```diff
@@ -5,7 +5,9 @@
 
 <%# BUTTON: Add in the deletion of all files %>
 <div class="btn-line">
-  <%= link_to "Delete All Files", polymorphic_path([main_app, :destroy_all_files, :hyrax, presenter.model_name.name.downcase.to_sym]), class: 'btn btn-danger', data: { confirm: "Are you sure you want to delete all files?" }, disabled: presenter.file_set_presenters.blank?, method: :delete %>
+  <% if presenter.file_set_presenters.blank? == false %>
```
straleyb, line 8: This works but instead of a comparison like this we can just do `if !presenter.file_set_presenters.blank?`

If it is blank itll return true, but then make it false so it wont fulfill the if statement.

If it isnt blank itll return false but then make the false true so itll fulfill the if statement.

straleyb (approved):

Issue #2442, opened 2023-03-17
Analytics pages not showing accurate or recent data


carakey:
### Descriptive summary

(1) There are major discrepancies between download counts pulled directly from of Google Analytics and those shown on an SA fileset's Analytics page, AND
(2) For most works the pageviews/downloads chart shown on the Analytics page cuts off too early. There is some variability between works, but few that are up to date -- I'm seeing charts ending in 2019, 2020, 2021, etc. 

Since the SA Analytics count doesn't match the GA count, it's hard to know if the total count is up to date or also ending when the chart stops.

Example: https://ir.library.oregonstate.edu/files/m613n196v/stats?locale=en

- SA Analytics page
  - Text data reads "0 views and 39592 downloads since October 2, 2012"
  - Chart ends at Oct 27, 2019
- Google Analytics downloads for this fileset
  - 409 total downloads
  - plenty of downloads since 10/27/19 (e.g. 4 downloads on 3/15/23)

### Expected behavior

Analytics counts and charts show accurate usage data from a work's creation to the present day, so that authors and campus units can rely on our metrics to demonstrate the impact of their work.

### Actual behavior

Analytics counts differ from Google Analytics data. Charts do not show data to present day.

### Steps to reproduce the behavior

Go to any work's fileset and click Analytics. Check last date shown on chart. Compare totals with Google Analytics

lamtu1:
@carakey  I was wondering can you give me a screenshot of the Google Analytic on the fileset that you put in the ticket as an example? I wanted to see what Google Analytic produce to compare to the one on SA. Thank you

lamtu1:
Thank you @carakey for the screen shot and it will a great comparison to look as trying to navigate and narrow the problem.

carakey:
@shieldsb @lamtu1 
This ticket is about questionable data in the SA / Hyrax analytics pages. We don't have those analytics pages currently, and haven't since the GA3 to GA4 cutover. I think this ticket should be "Icebox" and revisited after analytics are available in SA / Hyrax again. It's definitely not "Ready." 

I'll make a new ticket about re-establishing the user-facing analytics data.

PR #2477, opened 2023-05-04
Integrate the GA with API keys

lamtu1:
Add in the `Google Analytics` key to the YML file to see if this fix the `Scholars Archive` analytics.
#2442

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -1,9 +1,9 @@
 #
 # To integrate your app with Google Analytics, uncomment the lines below and add your API key information.
 #
-# analytics:
-#   app_name: GOOGLE_OAUTH_APP_NAME
-#   app_version: GOOGLE_OAUTH_APP_VERSION
-#   privkey_path: GOOGLE_OAUTH_PRIVATE_KEY_PATH
-#   privkey_secret: GOOGLE_OAUTH_PRIVATE_KEY_SECRET
-#   client_email: GOOGLE_OAUTH_CLIENT_EMAIL
+analytics:
+  app_name: <%= ENV.fetch('GOOGLE_OAUTH_APP_NAME', '') %>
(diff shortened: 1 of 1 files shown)


lamtu1:
@wickr I have gotten a chance to look at it, and Corey and I also sit down and talk about it. We look through `hyrax.rb` and thought it is connecting to GA via in `hyrax.rb` to send data on to GA from Scholars Archive. Because what we discover that GA doesn't automatically send data back to Scholars Archive to update it on SA end, we thought adding this back in might fix it. But I can run through with it with Corey once more on this situation.

wickr:
Ok, I see what you're saying. But if this file was never actually used, but the system of retrieving analytics data worked up until a bit ago, I would think it's something else.

CGillen:
I'm not convinced it was working until a little bit ago. @carakey @simholt any idea how long analytics have been out of sync?

As far as I can tell, the stats were synced from dspace ~6 years ago, then stopped updating around Jan 2022 for some reason.

From what I can trace, Hyrax attempts to pull analytics from GA, failing that they use `::FileDownloadStat` which I can't find where they're generated besides our `process_legacy_stats` rake task that did the dspace syncing. Comments to the analytics code claim `Hyrax::Download` objects are pushed to GA as events, I'm assuming for consumption later to build these analytics pages

wickr:
I'm guessing we lost another config or an API key (`.p12` file) somewhere along the way, or something got messed up with the database on our end.

Or, there was a Rake task running for stats, but it didn't get included in one of the system migrations. There might be helpful info here: https://github.com/uclibs/scholar_uc_legacy/issues/1767

simholt:
@CGillen No idea, they have been problematic off and on since the move to Hyrax.

carakey:
@CGillen All I can say for sure is that it was reported as busted by an OSU faculty on March 13.

CGillen (approved):

PR #2478, opened 2023-05-04
Add in the update to only login user can see it

lamtu1:
Add in a statement to make sure that the content will display only when user is log in.


CGillen:
This won't be enough to prevent viewing from students. Please use `presenter.editor?`

lamtu1:
Would the `presenter.editor?` check both if the user sign in and it knows who can delete the file or we would have to check that as well?

CGillen:
It'll check both. It's true iff the current person viewing the page can make edits to the work. IE: logged in and admin to the dataset admin set

CGillen (approved):

PR #2501, opened 2023-06-07
Update redirect_if_restricted_behavior.rb

straleyb:


CGillen:
This should be more generic:
```
URI(request.referrer).host == ENV.fetch('SCHOLARSARCHIVE_URL_HOST') && URI(request.referrer).path == '/'
```
We'll just need to set `SCHOLARSARCHIVE_URL_HOST=test.library.oregonstate.edu` on dev (no port)

CGillen (approved):

PR #2504, opened 2023-06-09
Create a tab for admin OCR

lamtu1:
Create a debug tab section showing off the OCR for admin.


app/views/hyrax/base/_guts4form.html.erb
```diff
@@ -0,0 +1,80 @@
...
+            <li role="presentation">
+          <% end %>
+              <a href="#<%= tab %>" aria-controls="<%= tab %>" role="tab" data-toggle="tab">
+                <%= t("hyrax.works.form.tab.#{tab}") %>
+              </a>
+            </li>
+        <% end %>
+
+        <% # NEW TAB: Add in a tab for OCR text display for admin %>
+        <% if f.object.model.persisted? && current_user.admin? && !f.object.model.file_sets.blank? %>
```
CGillen, line 28: Can you remove this override and use a system like this: https://github.com/osulp/Scholars-Archive/blob/main/app/helpers/hyrax/doi/work_form_helper.rb#L7-L12

You can add your method around here: https://github.com/osulp/Scholars-Archive/blob/main/app/helpers/hyrax_helper.rb#L24

As long as the value you insert is `ocr` your other view should just work

CGillen (changes requested):

app/views/hyrax/base/_guts4form.html.erb
lamtu1: Got it, I will try to work on changing the view. Also the code system you reference, I see that it do `super.insert(1, doi)`. And was wondering if that `1` is the index or insert 1 new value in?

CGillen: Correct, it's the index to insert into. So you can put it in a specific place or if you just want to put it on the end, use the shovel `<<`

CGillen (approved):

Issue #2505, opened 2023-06-13
Explanatory messaging for restricted/embargoed filesets


carakey:
### Descriptive summary

The desired behavior is to interrupt the automatic redirect to the login screen, and instead display a page with an explanation for filesets that are attached to restricted or embargoed works. OSU-only works include a login link that will put them through to the target fileset. This is for the same type of functionality as #2032 and #2494, only applied to filesets. 

### Expected behavior

When any user who is not logged in tries to access a restricted or embargoed fileset, they are presented with clear messaging that the work is embargoed / private / only available to OSU users. As applicable, OSU users understand that they can log in with their ONID and access it, while non-OSU users understand that they cannot access it. The login link will send them to the target fileset after they log in. This page is part of the ScholarsArchive system so they have other options to engage, such as home, help, and contact.

### Actual behavior

When any user who is not logged in tries to access a fileset with restricted visibility, they are redirected to the ONID login screen.

### Steps to reproduce the behavior

Attempt to access the fileset attached to any embargoed or permanently restricted work. Example https://ir-staging.library.oregonstate.edu/concern/parent/pk02c995f/file_sets/m613mx70x 

### Related work

#2032  
#2494

carakey:
The example fileset on staging, linked above, provided messaging and redirecting as intended. QA pass.

carakey:
Also reopening #2494 and #2032 for same reason

PR #2512, opened 2023-07-06
Add in redirect for fileset

lamtu1:
Add in the redirect to display message for the fileset that is under embargo.


CGillen:
https://github.com/osulp/Scholars-Archive/blob/main/app/controllers/application_controller.rb#L12
It looks like you were right to be digging around the base applicationcontroller. This might be overboard however. This will redirect every restricted item to the homepage & it includes a huge override for us to maintain.
We already have a monkey patch into the fileset controller, lets use that to narrow this down and not take on any more maintenance workload
https://github.com/osulp/Scholars-Archive/blob/main/config/initializers/hacks/hyrax_file_sets_controller_hack.rb

CGillen (changes requested):
https://github.com/osulp/Scholars-Archive/blob/main/app/controllers/application_controller.rb#L12
It looks like you were right to be digging around the base applicationcontroller. This might be overboard however. This will redirect every restricted item to the homepage & it includes a huge override for us to maintain.
We already have a monkey patch into the fileset controller, lets use that to narrow this down and not take on any more maintenance workload
https://github.com/osulp/Scholars-Archive/blob/main/config/initializers/hacks/hyrax_file_sets_controller_hack.rb

lamtu1:
@CGillen From what I understand is instead of working on the `controller.rb` and fixing the route in that, that would have a huge override for file sets. So rather than going to `controller.rb`, We already have the code fix for file sets in the `hack` you mention and use that to find a way to redirect with that show page in the hack?

CGillen:
Yes, `controller.rb` is a lot to maintain and check when we go for an update. the `file set controller hack` already allows you access to the fileset controller directly, so anything you write in there can be an override anywhere in the fileset controller hierarchy

CGillen (approved):

lamtu1:
fixes #2505

Issue #1493, opened 2018-07-16
Restrict Degree Name based on Degree Level for depositors


vantuyls:
For depositors, when a Degree Name is selected, they should only see Degree Levels that correspond to the Degree Name.

Degree Name should be "smart" based on what Degree Level is selected.

Level: Bachelor
Name:
- Bachelor of Science (B.S.)
- Bachelor of Arts (B.A.)
- Honors Bachelor of Arts (H.B.A.)
- Honors Bachelor of Fine Arts (H.B.F.A.)
- Honors Bachelor of Science (H.B.S.)

Level: Masters
Name: 
- Master of Agriculture (M.Ag.)
- Master of Arts (M.A.)
- Master of Arts in Interdisciplinary Studies (M.A.I.S.)
- Master of Business Administration (M.B.A.)
- Master of Education (Ed.M.)
- Master of Fine Arts (M.F.A.)
- Master of Ocean Engineering (M.Oc.E. )
- Master of Public Health (M.P.H.)
- Master of Science (M.S.)

Level: Doctoral
Name: 
- Doctor of Philosophy (Ph.D.)
- Doctor of Education (Ed.D.)

simholt:
This ticket needs to be brought up to date. The degree names are not current.

simholt:
@lamtu1 Okay! -- the new degree names (that were not listed in original ticket) are in BOLD. Having "Other" still be an option may lead to user error but it also helps us discover new programs and redirecting students.

Degree Level: Bachelor
Degree name options:
- Bachelor of Arts (B.A.) 
- **Bachelor of Fine Arts (B.F.A.)**
- Bachelor of Science (B.S.)
- Honors Bachelor of Arts (H.B.A.)
- Honors Bachelor of Fine Arts (H.B.F.A.)
- Honors Bachelor of Science (H.B.S.)
- **Other**

Degree Level: Masters
Degree name options:
- **Executive Master of Public Policy (E.M.P.P.)**
- Master of Agriculture (M.Ag.)
- Master of Arts (M.A.)
- Master of Arts in Interdisciplinary Studies (M.A.I.S.)
- Master of Business Administration (M.B.A.)
- Master of Education (Ed.M.)
- Master of Fine Arts (M.F.A.)
- **Master of Forestry (M.F.)**
- **Master of Natural Resources (M.N.R.)**
- Master of Ocean Engineering (M.Oc.E. )
- Master of Public Health (M.P.H.)
[...]

PR #2523, opened 2023-09-07
Update the degree sort

lamtu1:
Make changes to the `js` file that load the selection `on load` also rather than the `on change` only.


lamtu1:
fixes #1493

straleyb (approved):

Issue #1880, opened 2019-06-05
Rescue I18n::InvalidLocale


luisgreg99:
### Descriptive summary

When `locale` param is invalid, it raises `I18n::InvalidLocale`.

Application Trace:
```
app/controllers/concerns/hyrax/controller.rb:26:in `set_locale'

Request Parameters:
{"locale"=>"test"}

Response Headers:
None
```

### Expected behavior
It should render a 404 or a message that the parameter is invalid without raising the error.

### Actual behavior
It raises `I18n::InvalidLocale`

### Steps to reproduce the behavior

1. Go to http://localhost:3000
2. set locale to 'test': http://localhost:3000?locale=test

I think we could fix it by updating `set_locale` in https://github.com/samvera/hyrax/blob/master/app/controllers/concerns/hyrax/controller.rb#L26 and adding a method to validate the value there. If it fails with `InvalidLocale`, then rescue the error and display a message or maybe default to a valid value like `en`.

### Related work
https://github.com/samvera/hyrax/issues/2637

PR #2528, opened 2023-09-14
Add in error handling to locale

lamtu1:
Add in a way to handle the param for locale and not raise the `I18n::InvalidLocale`.


lamtu1:
`DEMO:` Showcasing the error handling of locale.

https://github.com/osulp/Scholars-Archive/assets/111306182/6df46475-f1b0-441d-82df-d044a7b4471f

lamtu1:
fixes #1880

lamtu1:
Ready for review/approve and merge for QA.

app/controllers/concerns/hyrax/controller.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+# OVERRIDE: Setup an override on locale to render header
+module Hyrax::Controller
```
CGillen, line 4: See that this is a module & not a class. We should be able to find where this is imported and override the method there without pulling in this whole module

CGillen: https://github.com/osulp/Scholars-Archive/blob/main/app/controllers/application_controller.rb#L12

CGillen (changes requested):

CGillen (approved):

Issue #1887, opened 2019-06-07
Remove Google Analytics calls from Admin Dashboard views


wickr:
### Descriptive summary

The Admin Dashboard has Google Analytics calls, so each pageview under there is getting tracked. I don't think it matters that `/admin/workflows?locale=en` was visited 33 times in the last week.

Confirm /notifications route/view GA calls are removed too, since it has its own route.

### Expected behavior

Admin Dashboard views do not need to be tracked for Google Analytics

### Actual behavior

Admin Dashboard views are being tracked for Google Analytics

### Steps to reproduce the behavior

1. Visit page in Dashboard
2. Look at Analytics > Realtime > Content view.

carakey:
I visited Staging admin tabs and then viewed the Google Analytics property for SA staging. GA confirmed that 1 user was active. I searched 'admin' in the Pages and Screens tab and got no results. Compare to searching for 'admin' in P&S on GA for prod and getting 6 hits on the `/admin/workflows` line. 

Tentative QA Pass, and then double check after it's in Prod.

PR #2538, opened 2023-10-17
Add in a filter to not collect admin visit

lamtu1:
Add in a special case to ignore some admin path visit to not be collected in the `GA4`.


lamtu1:
fixes #1887

straleyb (approved):

Issue #2532, opened 2023-09-21
Some other_affiliations not on show page or limiters


simholt:
### Descriptive summary
Newly assigned `other_affiliations` based on LC URIs are not displaying on show pages or under the Non-Academic Other limiter. ONS and Wikidata ones are fine. LC servers can be very busy, but this has been happening a lot and I can't find any pattern; it is us or them?

### Expected behavior
When a user selects a Non-Academic Other, it should be listed after the save on the show page and be collated with others under limiters.

### Actual behavior
`other_affiliations` with LC URIs don't show up, although they can be viewed if you go in to edit or if you pull the id with Solr. 
Example: https://ir.library.oregonstate.edu/concern/defaults/3r075360h has Hatfield listed as Non-Academic Other (LC URI: http://id.loc.gov/authorities/names/n84131837) but it is not on the show page.

I removed a Non-Academic Other from a work where it was shown correctly, then went in and added it again, and it did not show up.

Some other LC authorities:
http://id.loc.gov/authorities/names/n79115395
http://id.loc.gov/authorities/names/n79100852
http://id.loc.gov/authorities/names/n80126175

simholt:
### Steps to reproduce the behavior

1. Go into existing work and select Hatfield as a Non-Academic Other
2. Save
3. See Hatfield is not on show page
4. (Clean up and delete your addition!)

simholt:
I can add those URIs now but they are not resolving to the correct text on the show page:

**EDIT PAGE:**
![Screenshot 2023-11-07 at 3 32 56 PM](https://github.com/osulp/Scholars-Archive/assets/33705514/a6fcfb9c-2360-4fb1-82ad-e34fdbaae8a9)


**SHOW PAGE:** 
![Screenshot 2023-11-07 at 3 32 47 PM](https://github.com/osulp/Scholars-Archive/assets/33705514/dcf2b960-72b0-44d3-9377-104ed6a334e9)

lamtu1:
@simholt Would you like the show page to display the naming convention like in the `edit` form rather just a URL link? or would you want the naming convention to show up with the hyperlink to their specific affiliation?

simholt:
Cannot test until staging is sorted.

simholt:
So there are words showing up instead of URIs, but they are not the LC authorities. Example:`qv33rw93b`

[...]

PR #2546, opened 2023-11-08
Modify the other affiliation display

lamtu1:
Add in a fix in display to show the name of the  `other affiliation` rather than just the href link


lamtu1:
fixes #2532

straleyb (approved):

Issue #2548, opened 2023-12-07
Update OAI-PMH configuration


carakey:
### Descriptive summary

We are participating in a US Repository Network/SPARC/COAR pilot project to aggregate articles and metadata from IRs. It depends on OAI-PMH harvesting. The SA endpoint at https://ir.library.oregonstate.edu/catalog/oai failed the [OpenArchives validation](https://www.openarchives.org/Register/ValidateSite) used in the project. 

Failure message: `FAIL baseURL supplied 'https://ir.library.oregonstate.edu/catalog/oai' does not match the baseURL in the Identify response 'http://ir.library.oregonstate.edu/'. The baseURL you enter must EXACTLY match the baseURL returned in the Identify response. It must match in case (http://wibble.org/ does not match http://wibble.org/) and include any trailing slashes etc.` 

(It passed all other validation checks).

Per @CGillen:

> This looks like some kind of config on our end.
> https://ir.library.oregonstate.edu/catalog/oai?verb=Identify
> baseUrl uses http instead of https, which should be something we could change. 

### Expected behavior

OAI-PMH endpoint is configured such that the endpoint baseURL exactly matches that in the Identify response.  Endpoint passes validation at https://www.openarchives.org/Register/ValidateSite.

### Actual behavior

OAI-PMH endpoint fails validation with a non-matching baseURL.

### Steps to reproduce the behavior

Test the endpoint with the validator tool.

carakey:
This seems like something we can't fully QA and close until it makes it to Production and then passes validation. Looking at the changed file, it does appear to show a change from 'http' to 'https' as described. We can check again when it's in Prod.

carakey:
I ran the validation tool again and it still fails. It does show the update (http --> https) but otherwise still shows the same error message. 

![image](https://github.com/osulp/Scholars-Archive/assets/17655103/4a09baf8-4189-4cb5-82c6-7a2a97c801a4)

I'm reading this as, it expects the Identify response to return, literally, "https://ir.library.oregonstate.edu/catalog/oai".

lamtu1:
@carakey I think you are right, they want the `base URL` to match the entire thing rather than just the basic URL we have. So in this case, we would have to change the `base URL` case to have 
```
https://ir.library.oregonstate.edu/catalog/oai
```
rather than `https://ir.library.oregonstate.edu`

carakey:
It passed validation. 🎉 
  
[...]

PR #2550, opened 2024-01-02
Update OAI config from http to https

lamtu1:
`UPDATE:`  Add in the update to fix the `config.oai` to set the `basedURL` to reflect the `https` rather than the `http`.


lamtu1:
fixes #2548

CGillen (approved):

Issue #716, opened 2017-07-07
bibo:peerReviewed doesn't exist, look again at BIBO ontology


wickr:
`http://purl.org/ontology/bibo/peerReviewed` which is mapped in default metadata doesn't actually exist. 

There is a `http://purl.org/ontology/bibo/status/peerReviewed` and it's counterpart `http://purl.org/ontology/bibo/status/nonPeerReviewed` but they're term values for I think the bibo:DocumentStatus property.

The ontology is here: https://github.com/structureddynamics/Bibliographic-Ontology-BIBO/blob/master/bibo.owl

wickr:
@vantuyls @simholt 

Option A:
Change predicate to http://dbpedia.org/ontology/isPeerReviewed in DefaultMetadata and all items with this predicate
The values for that are of range xsd:boolean (one of 'true', 'false', 1, 0) so I think we're ok there, even though our TRUE/FALSE are capitalized

Option B:
Change predicate to http://purl.org/ontology/bibo/status in DefaultMetadata and all items with this predicate
Change PeerReviewed property values (TRUE/FALSE) to
* http://purl.org/ontology/bibo/status/peerReviewed
* http://purl.org/ontology/bibo/status/nonPeerReviewed

This would allow other Statuses defined by BIBO ('draft', 'forthcoming', 'legal', 'published', 'rejected, 'unpublished' have URIs) and any others we would want to define. The existing Peer Reviewed facet could be configured to stay as it is now, or it could be merged in to a bigger 'Document Status' one that has Peer Reviewed and Non Peer Reviewed along with the other values.

wickr:
Steve said go with Option A.

vantuyls:
@wickr please also update the MAP when you finalize this.

vantuyls:
@simholt @wickr if this is properly reflected on the MAP, please close this issue.

revgum:
The current thinking is that I run a job to change the predicate in Fedora in conjunction with a code update so the application knows about the property/predicate change, and our “QA” service is aware of the metadata value. The new values should be lowercase `true` and `false`. 

The solution to this ticket will need to be coordinated and planned to run the Fedora update immediately after the code is pushed to production.

carakey:
We are changing the predicate to `http://dbpedia.org/ontology/isPeerReviewed` (as in Option A, with agreement as documented in this ticket in Jan 2018). The new predicate should have xsd:boolean conformant values `true` or `false`.

Aspects of this change include:

[...]

PR #2552, opened 2024-01-05
Update Peer Review Ontology

lamtu1:
`UDPATE:` Make an adjustment to the ontology of the `peer review` subject term and updated to a new URI predicate and other areas that uses the old ontology.

### DIFF
--- a/lib/tasks/update_peer_reviewed_data.rake
+++ b/lib/tasks/update_peer_reviewed_data.rake
@@ -1,0 +1,61 @@
+namespace :scholars_archive do
+  desc "Update peer review predicate data from all work graphs & change value to lowercase"
+  task :update_peer_review_data, [:chunk_size] => :environment do |t, args|
+    # SETUP: Setup a default value of 100 for chunk_size
+    args.with_defaults(:chunk_size => 100)
+
+    # SETUP: Add in the start time of the bulk changes and add message to know that the rake is starting
+    datetime_today = Time.now.strftime('%Y%m%d%H%M%S') # "20171021125903"
+    Rails.logger.info "Processing bulk changes for peer review"
+
+    # ADD: Add in all the works that are in production to an array
+    Rails.logger.info "Finding all the works..."
+    all_records = [::AdministrativeReportOrPublication.all]
+    all_records << ::Article.all
+    all_records << ::BatchUploadItem.all
+    all_records << ::ConferenceProceedingsOrJournal.all
+    all_records << ::Dataset.all
+    all_records << ::Default.all
+    all_records << ::EescPublication.all
+    all_records << ::GraduateProject.all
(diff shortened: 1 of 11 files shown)


lamtu1:
fixes #716

lib/tasks/update_peer_reviewed_data.rake
```diff
@@ -0,0 +1,58 @@
...
+    all_records.each do |works|
+      works.each do |record|
+        # FETCH: Get the old peer review from the work
+        old_peer_review_statement = [record.resource.rdf_subject, old_predicate, nil]
+
+        # CONDITION: If the peer review section found in work, get the info out for preparing to be delete
+        if !record.resource.graph.query(old_peer_review_statement).statements.empty?
+          # GET: Get the graph
+          orm = Ldp::Orm.new(record.ldp_source)
+          new_update_value = record.peerreviewed.downcase
```
CGillen, line 39: Check this link for how to update with the previous value
https://github.com/osulp/Scholars-Archive/blob/main/lib/tasks/update_conference_section_data.rake#L29-L44

As this is I would expect it to always grab `nil`, since it's looking at the new predicate for `peerreviewed`

CGillen (changes requested):

lib/tasks/update_peer_reviewed_data.rake
lamtu1: Ah I see with this:
```
new_value = orm.query(old_conference_section_statement).first.object.value
```
it actually querying the old predicate to get the value out of it and assign it to the new predicate once the graph is deleted. So we should use this to map out to the old value rather just saying `fetch work which would hold the new peerreviewed` and it would contain `nil`

wickr:
The predicate here should remain `http`. Even though it's accessible by `https`, the dbpedia ontology and URIs are defined as `http`, see About: https://dbpedia.org/ontology/

CGillen (commented):
@lamtu1 this looks really good.
Can you match the update_conference_section_data task and make use of the `ReplacePredicateJob` job?
The chunking will help the number of jobs and it'll just give me that little bit of piece of mind since we've ran that code before

CGillen (changes requested):
Above comment

lamtu1:
Yes, I will look through and see if I could modify it in a way that we could use that `ReplacePredicateJob` to run the updating for the `peerreviewed`

CGillen (approved):

CGillen:
I'm going to hold merge until we have a new stable production deploy. I don't want too many changes going on at once and this is a pretty big one with a big job to run

lamtu1:
> I'm going to hold merge until we have a new stable production deploy. I don't want too many changes going on at once and this is a pretty big one with a big job to run

@CGillen I was wondering if we can merge this PR in now since the update to Hyrax 3.6 was stable

PR #2553, opened 2024-01-09
Revert back the original layout of Other Affiliation

lamtu1:
`CHANGES:` Revert the code back to the original state how it is in the beginning. Discover to be an issue of indexing rather than displaying itself.

Update: The `TripleStore_URL` was the main cause and got it updated to reflect the label to be able to pull.


lamtu1:
fixes #2532

CGillen (approved):

PR #2557, opened 2024-01-12
Change code to display affiliation

lamtu1:
`BUG:` Change one piece of code to accept `other affiliation` that is part of the `LOC` to display on the work show page.

### DIFF
--- a/app/services/scholars_archive/triple_powered_service.rb
+++ b/app/services/scholars_archive/triple_powered_service.rb
@@ -67,14 +67,12 @@ module ScholarsArchive
     def predicate_labels(graph)
       labels = {}
       return labels if graph.nil?
-
       rdf_label_predicates.each do |predicate|
         eng_labels = graph
           .query(predicate: predicate)
           .select { |statement| !statement.is_a?(Array) }
           .map(&:object)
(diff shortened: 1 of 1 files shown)


lamtu1:
fixes #2532

lamtu1:
`SCREENSHOT:` Showing the display on show page

Screenshot 2024-01-12 at 8 55 00 AM

app/services/scholars_archive/triple_powered_service.rb
```diff
@@ -73,15 +72,13 @@ def predicate_labels(graph)
           .query(predicate: predicate)
           .select { |statement| !statement.is_a?(Array) }
           .map(&:object)
-          .select { |value| value.respond_to?(:language) ? value.language.in?(%i[en en-us]) : true }
```
CGillen, line 76: We still want to prefer english labels if we can. So maybe we take the above `#map` call and save that as `labels[predicate.to_s] = all_labels.map(&:to_s)`
Then we can run the `#select` and see if it has anything in it. If it does, re-set `labels[predicate.to_s] = eng_labels.map(&:to_s)`?

CGillen (approved):

PR #2559, opened 2024-01-22
Update the OAI-PMH base URL

lamtu1:
`UPDATE:` Add in the `repository_url` to match with the `OAI` url presented to match exactly by character.


lamtu1:
fixes #2548

PR #2623, opened 2024-06-26
Initial setup for Bulkrax plus installation

straleyb:
One thing to note here is that there is some issues with docker-compose build I ran into. For some reason libxml-ruby wasn't installing because build-essential wasn't being installed properly. Im going to do some testing with how to change the Dockerfile to work with that. I think removing build-essential from the list of package installation and moving it onto its own line might fix it. It needs to be forced to allow the installation without confirmations.


straleyb:
Closed. I think Bulkrax is OK

Issue #2569, opened 2024-03-11
Retrieve information about SA's archive and torrent files


carakey:
For the file inventory that happens with the annual Preservation Assessment process, it would be great to have information about repository contents that I'm not able to get with the other filesets from Solr. These include the contents of archive/container files and the torrent files that are stored separately. I'm looking for basic characterization info: file name, format, size.

This is something I want to pull once a year. This could be either a workflow + instructions + access to retrieve the information at my convenience, OR a list of the information as of X date. 

The 2024 preservation assessment happens the last week of April. I can furnish a list of PIDs for container or torrent files if it helps.

lamtu1:
Hi @carakey, as looking through the ticket and on Scholars Archive, I have some questions hoping you could clarify more for me.

`1.` I was wondering would you want the report to be similar to the fixity report and send out an email with all the information listed above to run once a year?

`2.` On top of the report, would you like having just a button to give you the same report as the email mention above but you can use that button anytime you want to access it?

`3.` For the `archive/container files and the torrent files` do you know where we store that on SA?

Thank you!

carakey:
Hi @lamtu1. The first option, an email report once a year, would be enough for my needs. I would love to have the second option available (the button) but expect that's a lot more work. Ryan O can answer the location question if @CGillen can't - I don't know the directory info. Container files like ZIPs are uploaded to SA like any other file and stored in the same location, but I understand the torrents are on a separate server.

lamtu1:
Thank you for the information @carakey, I will go ask around and look around and get that rake task to do the emailing on those files

carakey:
(Following up from Slack)

The 2023 inventory turned up the following numbers by MIME type for torrent and archive files. Current IDs should be retrievable from Solr with this info. 

__Torrent files by `mime_type_ssi`, May 2023 counts:__

- application/x-bittorrent - 11

__Archive/Container files by `mime_type_ssi`, May 2023 counts:__

- application/zip - 549
- application/x-gzip - 12
- application/x-tar - 6
- application/x-7z-compressed - 2

lamtu1:
Hi @carakey, I have one more question I wanted to follow up with you. When you said ` I'm looking for basic characterization info: file name, format, size.`. Does this mean you are looking for `each file name, the format is it in, and the size of that file` that is inside for example the `zip` file? Below is an example:

`test.zip` has 2 items
-> File Name: `a.pdf`, format: `PDF`, size: `1KB`
[...]

PR #2627, opened 2024-07-02
Update the fileset check

lamtu1:
`BUGS:` Update the fileset check task to include if `:mime_type` is not present


lamtu1:
fixes #2569

PR #2640, opened 2024-07-16
Enqueue reindexing jobs to new reindex queue

CGillen:


config/initializers/sidekiq.rb
```diff
@@ -11,6 +11,23 @@
...
+      queue, count = queue_weighted.split(",")
+      count = count.to_i
+      [count, 1].max.times do
+        queues.push(queue)
+      end
+    end
+
+    s.options[:queues] = queues
+  else
+    s.options[:queues] = ["default","ingest","reindex"]
```
decimalator, line 29: do we want the default to include `reindex`?

PR #2651, opened 2024-08-08
Update other_affiliation.yml

simholt:
Adding Spring Poster Symposium.


CGillen:
See #2652 for why this was closed

Issue #2658, opened 2024-08-23
Geonames location not resolving on show page


simholt:
### Descriptive summary
You can add Geonames locations to a record and it saves fine (and shows up in limiters) but it doesn't appear on the show page. They were showing sporadically before the upgrade but now they don't show up at all.

### Expected behavior
Geonames location info listed on show page.

### Actual behavior
It's associated with the work but you can't see it unless you view the edit page.
Example: https://ir.library.oregonstate.edu/concern/defaults/qf85nk58b

### Steps to reproduce the behavior

1. Add something to a location field - like this one: `gt54kw13k` Save.
2. View the record -- nothing there. 
3. Click edit button and oh look the field is filled out.
4. Please remove your change!

lamtu1:
Hi @simholt, I just checkout the link you mention in this ticket, below is the screenshot on my end of what I see and it is displaying the `location` correctly.

![Image](https://github.com/user-attachments/assets/272d591b-70af-48d4-8713-7b8d92ec5d24)

And I also did a test on my `local` env, when I first created the work, it didn't appear and waited for couple minutes until it process and refresh the page, and the `location` appear for me (another screenshot below) and was wondering if it is not showing still for you on your end?

![Image](https://github.com/user-attachments/assets/8a8be0d1-897d-4a1b-8805-ce7892e6ecc0)

simholt:
[qf85nk58b](https://ir.library.oregonstate.edu/concern/defaults/qf85nk58b) That one shows for me now! I don't know what changed?

However, all the others I checked that have locations -- [like all these](https://ir.library.oregonstate.edu/catalog?f[based_near_label_sim][]=Klamath+Basin%2C+%2C+United+States%2C+(Area)&locale=en&q=&search_field=all_fields) -- the location is not showing up.
I tried re-saving it & waiting a while, deleting the location then re-entering, and still nothing.

lamtu1:
Hmm I see it too, I will have to dive deep into it and see what I can do

lamtu1:
`UPDATE:` Discover that the `geoname location` is using `sidekiq worker` to perform solr indexing which it will take a couple minutes (roughly 1-2 min) to be index and the display will show up after waiting a couple minutes and refresh the page to see the change.

simholt:
@lamtu1 Q: So it sounds like we will need a reindex for locations to show up on existing records?



[...]

PR #2668, opened 2024-09-05
Modify the display of the location geoname

lamtu1:
`UX:` Fix the display based on the request from ticket for `geoname` location.


lamtu1:
fixes #2658

lamtu1:
`Image:` Display of the `show` and `search` page

Screenshot 2024-09-05 at 3 46 21 PM
Screenshot 2024-09-05 at 3 46 56 PM

app/controllers/catalog_controller.rb
```diff
@@ -97,7 +97,7 @@ def self.modified_field
     config.add_facet_field 'has_journal_sfacet', limit: 5, label: 'Journal Title'
     config.add_facet_field 'language_label_ssim', label: 'Language', limit: 5
     config.add_facet_field 'license_label_ssim', label: 'License', limit: 5
-    config.add_facet_field solr_name('based_near_label', :facetable), label: 'Location', limit: 5
```
straleyb, line 100: Why are we changing this name? Could we just keep the name but change what is stored in the field?

app/controllers/catalog_controller.rb
```diff
@@ -203,7 +203,7 @@ def self.modified_field
...
+        qf: "#{all_names} title_tesim alternative_title_tesim contributor_advisor_tesim contributor_committeemember_tesim abstract_tesim dspace_community_tesim dspace_collection_tesim degree_grantors_label_tesim nested_related_items_label_tesim nested_ordered_creator_label_tesim nested_ordered_additional_information_label_tesim nested_ordered_contributor_label_tesim nested_ordered_abstract_label_tesim degree_field_label_tesim file_format_tesim all_text_tsimv language_label_tesim rights_statement_label_tesim license_label_tesim academic_affiliation_label_tesim other_affiliation_label_tesim based_near_parse_label_tesim based_near_parse_label_tesim web_of_science_uid_tesim",
```
straleyb, line 206: based_near_parse_label_tesim showing up twice here.

app/models/solr_document.rb
```diff
@@ -82,7 +82,7 @@ def based_near_linked
   end
 
   def based_near_linked_label
-    ScholarsArchive::LabelParserService.parse_label_uris(self['based_near_linked_ssim'])
+    ScholarsArchive::LocationLabelParserService.location_parse_uris(ScholarsArchive::LabelParserService.parse_label_uris(self['based_near_linked_ssim']))
```
straleyb, line 85: If ScholarsArchive::LabelParseService needs more functionality, id rather that exist in the ::LabelParseService rather than create a new object that does the work. Unless im missing something about how ::LabelParseService needs to still function like it is, but in this one instance needs to be extended.

straleyb (commented):
So mostly just confused about the purpose of the new solr_name and facet field changes to those new names.

straleyb (approved):

PR #2671, opened 2024-09-09
Modify the parser a bit for geoname

lamtu1:
`UPDATE:` Modify the function of the parser to make sure the label parse correct value.


lamtu1:
fixes #2658

Issue #2738, opened 2025-03-12
Funding Body field duplicated on edit page


simholt:
### Descriptive summary
This only shows up when editing existing pages:

<img width="641" alt="Image" src="https://github.com/user-attachments/assets/c5b1482f-c566-4195-9028-56ccd5fede8e" />

### Expected behavior
Only one field!

PR #2739, opened 2025-03-13
Remove an extra field in funding body

lamtu1:
`BUG:` Add in a way to remove an extra field on `:new` or `work have blank` funding body for work on depositor form.


lamtu1:
fixes #2738

PR #2768, opened 2025-05-07
Adds in rudimentary turnstile on all pages

straleyb:


app/controllers/bot_detection_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  # key in rack env that says challenge is required
+  class_attribute :env_challenge_trigger_key, default: 'bot_detect.should_challenge'
+
+  def self.bot_detection_enforce_filter(controller)
+    return unless enabled && !controller.session[session_passed_key].try { |date| Time.new(date) < session_passed_good_for }
+
+    return unless controller.request.get?
+
+    Rails.logger.info 'Redirecting for Turnstile'
+    controller.redirect_to '/challenge', status: 307
```
CGillen, line 31: Without the `dest` param there wouldn't be anywhere to redirect back to in the JS

app/views/bot_detection/challenge.html.erb
```diff
@@ -0,0 +1,61 @@
...
+        });
+  
+        if (!response.ok) {
+          throw new Error('bad response: ' + response.status + ": " + response.url);
+        }
+  
+        result = await response.json();
+        if (result["success"] == true) {
+          const dest = new URLSearchParams(window.location.search).get("dest");
+          window.location.replace("/catalog");
```
CGillen, line 42: I think we should still use the `dest` param, incase this is a saved search. Sounds like that's not uncommon

CGillen (changes requested):

app/views/bot_detection/challenge.html.erb
straleyb: Totally. I put this in here *To Just Get It Working* because i wanted it at least working lol. It was doing something funky with the destination Url that it was pulling. It was coming up as nil and I didn't really look into why. I saw it was replacing the subdirectory with whatever was nil. So I just dropped catalog in there.

So yeah, this implementation only works for catalog at the moment.

Do you think its worth while to get this in at least and get it tested and onto prod while I extend it?

CGillen: Sure, if you wanna start getting this tested that's fine

PR #2769, opened 2025-05-09
Makes redirect based on initial url

straleyb:


CGillen (approved):
Tentatively approved, awaiting moving out of draft

PR #2784, opened 2025-06-10
Initial pass through for External Dataset

straleyb:


app/controllers/concerns/scholars_archive/datasets_controller_behavior.rb
```diff
@@ -30,8 +30,14 @@ def check_human_data(file_check)
 
     # METHOD: Get the value out of the human data and remove it
     def find_remove_human_data
-      get_data = params['dataset']['human_data']
-      params['dataset'].delete('human_data')
+      get_data = ''
+      if curation_concern.class.to_s.include?('External')
```
CGillen, line 34: This can probably generalize with `#underscore`

```
class_name = curation_concern.class.to_s.underscore

get_data = params[class_name]['human_data']
params[class_name].delete('human_data')
```

```
irb(main):006:0> "Dataset".underscore
=> "dataset"
irb(main):007:0> "ExternalDataset".underscore
=> "external_dataset"
```

CGillen (changes requested):
There's a `logs.txt` at root that sneaked its way in

straleyb:
Got the #remove_human_data method generalized and removed the logs.txt debug file.

PR #2804, opened 2025-08-21
Hard code urls to work for redirect monkey business

straleyb:
This hardcodes in the urls to work. we didnt have this working for staging and it was bugged locally.

On production, I couldn't get the actual environment variable to work properly. It should have been outputting properly but it wasn't. 

Checking the rancher config map the output should have been ir.library.oregonstate.edu but it wasn't finding the equivalency. 

Removing and abstracting to a list allowed me to extend it to all the other environments as well with one guard statement instead of 3.

redirect-test

### DIFF
--- a/app/controllers/concerns/scholars_archive/redirect_if_restricted_behavior.rb
+++ b/app/controllers/concerns/scholars_archive/redirect_if_restricted_behavior.rb
@@ -14,10 +14,11 @@ module ScholarsArchive
       def redirect_if_restricted
         curation_concern = ActiveFedora::Base.find(params[:id])
         referrer_uri = URI.parse(request.referrer.to_s)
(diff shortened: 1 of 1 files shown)


CGillen:
What about something more like this to test if we're referred from the homepage?
```
referrer_uri = URI.parse(request.referrer.to_s)
referred_from_homepage = referrer_uri.host == ENV.fetch('SCHOLARSARCHIVE_URL_HOST') && referrer_uri.path == '/'
```

I also noticed `SCHOLARSARCHIVE_URL_HOST` isn't set in our `docker-compose.yml` by default, so that could have been causing some problems in local development.

straleyb:
When I change the ENV variable to a hard coded host `ir.library.oregonstate.edu` it works. but ENV.fetch('SCHOLARSARCHIVE_URL_HOST') doesnt. 

Screenshot 2025-08-22 at 8 06 22 AM
ConfigMap does indeed look correct

Issue #2056, opened 2020-02-26
Remove rubocop_errors.txt


straleyb:

CGillen:
Referring to `rubocop_fix_me.yml`

PR #2816, opened 2025-09-10
Fix some small rubocops from the YML file

lamtu1:
`ADD:` Fix some small single rubocop fix me in the YML file.


lamtu1:
fixes #2056

PR #2830, opened 2025-10-14
Domain allow list

straleyb:


CGillen (approved):
We'll probably want this an ENV var later during the refactor

PR #2888, opened 2026-04-14
Updates to check persisted

straleyb:
This makes sure that on New form, the attestation isn't chosen.

On Edit, since the user already chose their attestation, they should already be in contact with the person for help, otherwise their work is accessible. So now we default to no help needed.

Do we want to actually persist this data? We could extend functionality further but that would mean adding the data to the object which would be heavy changes.


simholt:
@straleyb I don't think we need to save that data.

Issue #2752, opened 2025-03-24
v4 Functionality Issues


CGillen:
### Descriptive summary
This will be a running ticket of functional issues for the Hyrax v4 update.

Try to keep it to one issue per comment & provide a screenshot or link w/ reproduction steps

Some testing ideas:
- Create a work
- Edit a work
- Version a fileset
- Review a work through each workflow
- Create a collection
- Edit a collection
- Add work(s) to a collection
- Embargo/Lease stuff
- Examine all pages for load & content (even if it looks funky)
- Facet a search
- Use a facet from the facet modal
- Try paging through pagers
- Advanced search
- Open all the modals
- Spotlight work/researcher
- Other things you do on a frequent basis?

EDIT: Most of the styling has been updated, if you see something that looks wrong or the colors/sizing/shape/position/etc is off, comment it

### Related work

https://ir-testv4.library.oregonstate.edu/help?locale=en

simholt:
- [x] Cannot add new works. 

I click Share your work and the Select type of work window shows but then I get:

<img width="1135" height="545" alt="Image" src="https://github.com/user-attachments/assets/704fa519-b985-47a2-b638-018a42eae72e" />

simholt:
Cannot view existing works -- clicked through items listed on the Works page and on the catalog page. Get same message as above.

simholt:
- [x] Cannot upload file when creating a new work.

File looks like it is uploading but then says "Internal Server Error". When you click Submit, you get the broken message.

<img width="522" height="261" alt="Image" src="https://github.com/user-attachments/assets/81bab69f-d671-45ab-8f54-e2419828c319" />

simholt:
- [x] Ordered lists - duplicate problem
**Title
Creator
Contributor
Abstract
Related Items
Additional Information**

If the field has text when you click Add another, it is repeated:

<img width="602" height="320" alt="Image" src="https://github.com/user-attachments/assets/1e445fec-e18e-4a53-8598-c8dfe8de1f30" />

simholt:
- [x] Ordered list - no "cannot add another with empty field" note
**Title
Creator
Contributor
Abstract
Related Items
Additional Information**

If the field is blank when you click Add another, a blank field is added instead of the block:

<img width="487" height="402" alt="Image" src="https://github.com/user-attachments/assets/9be5caca-1012-433a-beef-b05d92e88eb2" />

Correct:

[...]

PR #2891, opened 2026-04-20
Fix RDF graph query call

CGillen:
Should fix this:
https://github.com/osulp/Scholars-Archive/issues/2752#issuecomment-4246321044


app/services/scholars_archive/triple_powered_service.rb
```diff
@@ -7,8 +7,10 @@ module ScholarsArchive
   class TriplePoweredService
     def fetch_all_labels(uris)
       labels = []
+      Rails.logger.info('###')
       uris.each do |uri|
         graph = fetch_from_store(uri)
+        Rails.logger.info(uri)
```
straleyb: Same as above

Issue #2568, opened 2024-03-11
Fixity email report fields aren't pulling correct data


carakey:
### Descriptive summary

Look into how the fixity checker is counting and reporting passing / failing files in its data fields. The last report had a discrepancy where it says zero files failed, but gives 14 IDs of failed files.

### Expected behavior

Counts of passing and failing files are reported accurately in the fixity report email.

### Actual behavior

The following is the fixity report emailed on on Feb 17, 2024:

> Fixity Report
> Hi there, below are the report on the Fixity Job that ran and the results after it has finished running:
> 
> Start Time: 2024-02-16 15:48:47 -0800
> End Time: 2024-02-17 01:03:08 -0800
> No. of Files [CHECKED]: 157268
> No. of Files [PASSED]: 157268
> No. of Files [FAILED]: 0
> List of all Failed Files [IDs]:
> 
> [kp78gg577](https://github.com/concern/file_sets/kp78gg577) 
> [dj52wc82j](https://github.com/concern/file_sets/dj52wc82j) 
> [41687r184](https://github.com/concern/file_sets/41687r184) 
> [br86bb654](https://github.com/concern/file_sets/br86bb654) 
> [qn59qb39p](https://github.com/concern/file_sets/qn59qb39p) 
> [8c97kx72p](https://github.com/concern/file_sets/8c97kx72p) 
> [1z40m1672](https://github.com/concern/file_sets/1z40m1672) 
> [qv33s449d](https://github.com/concern/file_sets/qv33s449d) 
> [pc289r65p](https://github.com/concern/file_sets/pc289r65p) 
> [jm214w89m](https://github.com/concern/file_sets/jm214w89m) 
> [v979v971k](https://github.com/concern/file_sets/v979v971k) 
> [vh53x358h](https://github.com/concern/file_sets/vh53x358h) 
> [9p290h89n](https://github.com/concern/file_sets/9p290h89n) 
> [hx11xp80b](https://github.com/concern/file_sets/hx11xp80b) 
> 
> - Scholars Archive Support
> 
> Email Date & Time: February 17, 2024 - 1:03 AM

I'm reading this as "100% of files passed. Zero files failed. These 14 files failed." If the last thing is true, then the first two things are not. 

### Steps to reproduce the behavior

Fixity report emails are saved in ScholarsArchive email archive.

### Related work

#2551 ; #2421

carakey:
I don't have a way to QA without seeing the monthly email report.

[...]

PR #2911, opened 2026-07-17
Fix the fixity count to include passed & failed

lamtu1:
`UPDATE:` Update the minor code to include in the checked passed & failed in `Fixity`.


lamtu1:
fixes #2568

straleyb (approved):

Issue #2316, opened 2021-12-21
Add favorite collections to dashboard


simholt:
### Descriptive summary
Requested by fac/staff who need to monitor deposits per graduation requirements (ex: Doctor of Veterinary Medicine) -- it would be helpful if the associated collection could be marked as a favorite on the user's SA Dashboard, even if they have not created it or contributed to it.

I think the dashboard needs an overhaul regardless but this would be helpful to a number of people.

lamtu1:
Anyone can QA this once finish building and push to staging.

straleyb:
@carakey This one is ready for QA.

carakey:
QA fail

I do see the new Favorite Collections tab and am able to add favorite collections from the dashboard lists of collections. 
However, a non-admin user can only view their own collections from the dashboard, so they cannot add collections to their favorites that they do not own. 

Could we add the Favorite star on the public view of a collection?

lamtu1:
Yes i can do that and will work on non-admin as well.

carakey:
@lamtu1 To be clear, non-admins do not need to access All Collections from the dashboard, but they do need a way to add favorites that they don't own. Having the option from the public view should take care of this need.

lamtu1:
@carakey Got it, would this mean you want to add a favorite button to the public view (show page) of the collection, and a non-member can add it via there?

carakey:
Yes exactly

lamtu1:
Will do, thanks for the clarification.

carakey:
Ultimately (I believe) the purpose of this feature is to let the user get to the public view of the collection, not the management of the collection. So they can see new works posted, etc. But NOT edit the collection object.

KennaW:
Per zoom with Cara and Tu. Favorites are visible. I can add my own collections and others collections.

The title of the "User Collection" ("[no toenails either](https://ir-staging.library.oregonstate.edu/dashboard/collections/x346d4386?locale=en)") goes to 'page not found'.

<img width="1569" height="729" alt="Image" src="https://github.com/user-attachments/assets/5492df8e-a1a4-44d7-9d20-e6f0a1d5e56f" />

carakey:
QA Pass. 🎉 

- Favorite button appears on public show pages for collections and functions as expected -- favorite/unfavorite adds or removes collections from favorites list. 
- Tested with a non-admin user (thank you Kenna) and verified that: 
    - non-admins see the Favorite Collections tab on their dashboard
[...]

PR #2924, opened 2026-07-28
Ad in minor code for linking to show page

lamtu1:
`UPDATE:` Add in two different links depend if they can edit the collection


lamtu1:
fixes #2316

Issue #2930, opened 2026-08-10
500 error on work


carakey:
### Descriptive summary

The work `cz30q315g` (https://ir.library.oregonstate.edu/concern/articles/cz30q315g/) is giving a 500 error. I am 95% certain it was loading fine last week and I have not made new edits. 

It's a slightly odd work because it is a compound object parent that also has a fileset - in other words, it has 2 member objects, one being a fileset and one being an Article work (which then has its own fileset).

There's some urgency here because the work is needed for a review by the ISSN Center, looking to assign an ISSN to the publication series in SA.

lamtu1:
@carakey I was wondering, does this work is attach to another work?

lamtu1:
I was able to access the edit page, but I think I know the culprit for the show page if this is the case

carakey:
@lamtu1 I'm not sure I understand the question, beyond what I already said at the top. This is the parent object in a compound object work. It has another work as a child object. It also has a fileset as a child object.

lamtu1:
@carakey That was it, I just wanted to make sure that's all, sorry for the double asking on it. But I know where the bug lies and will get a fix of it

lamtu1:
@carakey It is up on staging, can you test to see if you add a compound object to a work, will that work render the show page? if so, we can pass this and push to production

carakey:
@lamtu1 Just making sure we're clear: You mean add a work to another work, thus creating a compound object, right?

lamtu1:
@carakey Yes that what I mean, thank you!

carakey:
QA pass

QA process:

- Created a [parent work](https://ir-staging.library.oregonstate.edu/concern/conference_proceedings_or_journals/0r9673976) (Journal or Conference Proceeding) with file uploaded. Used a JPG image file for testing. 
- From saved but not-yet-approved parent work, clicked "Attach Child" >> "Attach Article" and created a [child work](https://ir-staging.library.oregonstate.edu/concern/articles/df65v814x) with file uploaded. Used a JPG image file for testing. 
- Approved the parent work; child work (as an Article) is automatically public / not routed for approval.
- In separate incognito tab, not signed in, attempted to view both parent work and child work using direct links to work pages.
- Both pages loaded.
- However the child work did not display the image. Troubleshooting:
    - Deleted the child image file.
    - Edited the child work to add a different file and update metadata (source is important!)
[...]

PR #2931, opened 2026-08-11
Change only fileset get display

lamtu1:
`UPDATE:` Fix the a bug on show page that deals with oembed with compound object.


lamtu1:
fixes #2930

CGillen (approved):

